PR #2, opened 2021-03-09
Adjust `maximum_bytes_per` to `maximum_bytes_per_file`

blt:
Based on @jszwedko's feedback the former name of this configuration
option was confusing. Not good to be confusing when we can help it.


jszwedko (approved):
Awesome, thanks!

PR #67, opened 2021-07-23
Introduce a new generator: kafka

### LINKED ISSUE #65: Make `kafka_gen`
Files and HTTP endpoints shouldn't have all the fun!

tobz:
Closes #65 

Introduces a new generator -- `kafka_gen` -- which produces messages to Kafka.  This first cut does not address any sort of parallelization of producers, since we should be able to hit modest throughput rates even with a single producer task.

We've added support to precache blocks, can generate all of the existing data variants, etc etc, and allow for multiple producer targets, similar to `http_gen`, which maps, essentially, to a specific Kafka server (cluster, really) and topic combination.

### DIFF
--- a/lading_generators/src/kafka_gen/worker.rs
+++ b/lading_generators/src/kafka_gen/worker.rs
@@ -9,6 +9,7 @@ use rdkafka::config::FromClientConfig;
 use rdkafka::producer::{FutureProducer, FutureRecord};
 use rdkafka::util::Timeout;
 use rdkafka::ClientConfig;
+use std::convert::TryInto;
 use std::num::NonZeroU32;
 
 #[derive(Debug)]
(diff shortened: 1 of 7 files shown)


lading_generators/src/kafka_gen/worker.rs
```diff
@@ -0,0 +1,216 @@
...
+    ///
+    /// Function will panic if user has passed non-zero values for any byte
+    /// values. Sharp corners.
+    #[allow(clippy::cast_possible_truncation)]
+    pub fn new(name: String, target: Target) -> Result<Self, Error> {
+        let mut rng = rand::thread_rng();
+
+        let rate_limiter = match target.throughput {
+            Throughput::Unlimited => unsafe {
+                // SAFETY: This comment is just here so I don't get fired.
```
blt, line 89: lol

lading_generators/src/kafka_gen/worker.rs
```diff
@@ -0,0 +1,216 @@
...
+            ),
+            Variant::Json => {
+                construct_block_cache(&payload::Json::default(), &block_chunks, &metric_labels)
+            }
+            Variant::FoundationDb => construct_block_cache(
+                &payload::FoundationDb::default(),
+                &block_chunks,
+                &metric_labels,
+            ),
+            Variant::Static { .. } => unimplemented!(),
```
blt, line 134: Intentional? I'd be happy to leave `static` out if we don't want it for this tool. TBH it's barely been useful for `file_gen`.

blt (approved):

lading_generators/src/kafka_gen/worker.rs
tobz: Oh, good point.. I just copypasted, but I should just not have variant.

blt (approved):

PR #70, opened 2021-07-27
Add kafka_gen to release workflow

tobz:
As stated in the title.


blt:
Oops, looks like we're also missing `http_blackhole`. I'd appreciate if you could toss that in too.

blt (approved):

PR #72, opened 2021-07-29
Give Datadog payload more structure

blt:
This commit gives the Datadog payload more intentional structure. Previously it
was very much the output of a fuzzing engine -- which lading basically has
embedded -- but this made any introspection on the payload quite challenging
needlessly.


spencergilbert (approved):
aside from clippy, lgtm

PR #77, opened 2021-08-05
Introduce `tcp_gen`, fluent payload

blt:
This PR introduces a new generator for TCP and a new payload, fluent. The TCP generator at present only supports fluent payloads, though this is easy to modify in the future if we desire. I have not included the wide-open bytes-per-second configuration option that `kafka_gen` has, though I think we probably do want to spread that across all the generators. As with `http_gen` the `tcp_gen` will keep pace with its target and unlike `http_gen` only has a single connection at present.

### DIFF
--- a/lading_common/src/payload/fluent.rs
+++ b/lading_common/src/payload/fluent.rs
@@ -18,7 +18,7 @@ enum RecordValue {
     Object(HashMap<String, u8>),
 }
 
-fn object<'a>(u: &mut Unstructured<'a>) -> arbitrary::Result<RecordValue> {
+fn object(u: &mut Unstructured<'_>) -> arbitrary::Result<RecordValue> {
     let mut obj = HashMap::new();
     for _ in 0..u.arbitrary_len::<(AsciiStr, u8)>()? {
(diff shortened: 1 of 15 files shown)


lading_common/src/payload/fluent.rs
```diff
@@ -0,0 +1,241 @@
...
+    Forward(FluentForward),
+}
+
+impl<'a> arbitrary::Arbitrary<'a> for Member {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let choice = u.arbitrary::<u8>()?;
+        let res = match choice % 2 {
+            0 => Member::Message(u.arbitrary::<FluentMessage>()?),
+            1 => Member::Forward(u.arbitrary::<FluentForward>()?),
+            _ => unreachable!(),
```
jszwedko, line 157: Just noting that the motivator for this use-case seems to be using fluentbit with a configuration, close to the default, that seems to always send `Forward` frames.

FluentD, on the other hand, seems to default to sending `CompressedPackedForward` frames. We may want to add support for that later.

PR #109, opened 2021-11-10
Add Splunk HEC blackhole

001wwang:
This PR adds a simple Splunk HEC blackhole that handles the following Splunk HEC paths
- POST `/services/collector/event | /services/collector/event/1.0 | /services/collector/raw | /services/collector/raw/1.0`
```jsonc
// Example response
{
  text: "Success",
  code: 0,
  ackId: 0,
}
```
`ackId` is a global `AtomicU64` that increments after each request, wrapping to 0 if needed. For simplicity, there is no notion of Splunk channels in this blackhole, but clients using channels can still successfully interact with this server. There is no Splunk HEC protocol guarantee that each channel receives `ackId`'s starting from `0`. 
- POST `/services/collector/ack`
```jsonc
// Example response
{
  acks: {
    0: true,
    1: true,
  }
}
```
For any ack query, we map over the provided `ackId`'s and return true.

- Requests to any other path result in a 200 OK with no body.

### DIFF
--- a/lading_blackholes/src/bin/splunk_hec_blackhole.rs
+++ b/lading_blackholes/src/bin/splunk_hec_blackhole.rs
@@ -1,4 +1,10 @@
-use std::{collections::HashMap, fs::read_to_string, net::SocketAddr, sync::atomic::{AtomicU64, Ordering}, time::Duration};
+use std::{
+    collections::HashMap,
+    fs::read_to_string,
+    net::SocketAddr,
+    sync::atomic::{AtomicU64, Ordering},
+    time::Duration,
+};
 
 use argh::FromArgs;
 use hyper::{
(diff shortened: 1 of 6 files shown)


lading_blackholes/src/bin/splunk_hec_blackhole.rs
```diff
@@ -0,0 +1,191 @@
...
+                }
+                Err(_) => {
+                    *okay.status_mut() = StatusCode::BAD_REQUEST;
+                }
+            }
+        }
+        _ => {}
+    }
+
+    Ok(okay)
```
blt, line 142: I dunno why but `Ok(okay)` made me giggle.

blt (approved):
Looks excellent.

PR #142, opened 2022-01-21
Copy `tcp_blackhole` binary to Docker container

pablosichert:
Follow-up to https://github.com/blt/lading/pull/139.


spencergilbert:
We should update the dockerfile to not need these hard coded and rather just grab all of the executables...

spencergilbert (approved):

PR #160, opened 2022-03-29
Allow introspection of target process

blt:
This commit introduce an 'introspection' sub-process, allowing the user to run 
any additional program to introspect the target process. Consider this config:

```
inspector:
  command: sh
  arguments:
    - "-c"
    - "perf record --verbose --compression-level=22 --call-graph=dwarf --pid=`pidof vector` --output=/tmp/vector.perf.data"
  environment_variables: {}
  output:
    stderr: "/tmp/captures/inspector.stderr.log"
    stdout: "/tmp/captures/inspector.stdout.log"
```

Lading'll manage the sh, the sh'll manage perf and perf'll record captures. This
allows perf to correctly shut down, while giving the user way more
freedom. We don't pass the PID of the vector sub-process yet, which is a pain, 
but is something potentially resolvable if it ends up being a problem. 

The biggest change here has to do with shutdown. It's now possible to clone
`Shutdown` without reference to any kind of channel, meaning we can thread a 
`Shutdown` through as many sub-tasks as we need, REF comments on #156. 


blt:
I think this is overly specific. Having an `analysis` sub-target in the configuration makes more sense to me now that I've had some time to let this sit. Allow the user to specify a path to a shell script or what not: `path` + `args` + environment variables?

blt:
> I think this is overly specific. Having an `analysis` sub-target in the configuration makes more sense to me now that I've had some time to let this sit. Allow the user to specify a path to a shell script or what not: `path` + `args` + environment variables?

Corrected with [a89637c](https://github.com/blt/lading/pull/160/commits/a89637c15fe9f49b37cbe9648e127aa0ea195c32)

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.58.1"
+channel = "1.60.0"
```
blt, line 2: This is unrelated to the change. I'll issue another PR to do the bump.

src/target.rs
```diff
@@ -103,56 +48,64 @@ impl Server {
...
     /// # Panics
     ///
     /// None are known.
-    pub async fn run(mut self) -> Result<(), Error> {
-        let mut child = self.command.spawn().map_err(Error::Io)?;
-        let wait = child.wait();
+    pub async fn run(mut self) -> Result<ExitStatus, Error> {
+        let config = self.config;
+        let mut target_child = match config.command {
+            Cmd::Path(p) => {
```
blt, line 71: Could probably collapse and remove this enum.

blt:
Oops, accidentally closed this with #167. Apparently you can't _just_ close a comment.

rust-toolchain.toml
blt: Addressed in #167

PR #199, opened 2022-05-16
Pass target pid as environment variable TARGET_PID

blt:
This commit passes the target pid as TARGET_PID to the inspector. This was
easier to deal with than an argument, as it allows the called program to do
whatever they like in terms of arguments.


blt:
This is pretty good, but would be more useful with an enable/disable flag. That way the inspector could always be flagged on in config and we can adjust at runtime whether the inspector actually impacts runtime.

vladimir-dd (approved):

PR #200, opened 2022-05-17
Collect CPU, memory, scheduling information about the target process

blt:
This commit introduces procfs dependency to the project and uses it to power an
'observer' server that runs as a peer to the target etc. This observer is
responsible for periodically reading the /proc/PID/stat of the target process,
writing out interesting details regarding memory consumption, CPU and scheduling
time. In this commit we collect:

* page_size -- the memory page size of the OS in bytes
* ticks_per_second -- the OS ticks that happen in a second
* rss_pages -- the real memory pages the process has allocated
* rsslim_bytes -- the soft limit of RSS in bytes
* num_threads -- the number of threads the process has running
* utime_ticks -- the number of ticks the process has spent scheduled in user-space
* stime_ticks -- the number of ticks the process has spent scheduled in kernel-space
* vsize_bytes -- the size in bytes of the process' virtual memory

### DIFF
--- a/src/observer.rs
+++ b/src/observer.rs
@@ -100,11 +100,11 @@ impl Server {
         let ticks_per_second: u64 = procfs::ticks_per_second()
             .expect("cannot determine ticks per second")
             .try_into()
-            .unwrap();
+            .expect("cannot convert ticks to u64");
         let page_size: u64 = procfs::page_size()
             .expect("cannot determinte page size")
             .try_into()
-            .unwrap();
+            .expect("cannot convert page size to u64");
 
         gauge!("page_size", page_size as f64);
         gauge!("ticks_per_second", ticks_per_second as f64);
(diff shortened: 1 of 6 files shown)


src/observer.rs
```diff
@@ -0,0 +1,145 @@
...
+            .expect("target failed to transmit PID, catastrophic failure");
+        drop(pid_snd);
+
+        let process = Process::new(target_pid.try_into().expect("PID coercion failed"))
+            .map_err(Error::ProcError)?;
+
+        let ticks_per_second: u64 = procfs::ticks_per_second()
+            .expect("cannot determine ticks per second")
+            .try_into()
+            .unwrap();
```
jszwedko, line 103: Should we do `expect`s here too for the integer conversion?

jszwedko (approved):
Nifty!

src/observer.rs
blt: Ah yeah, good call.

blt: Addressed in 503e73a.

Issue #155, opened 2022-03-23
`splunk_hec` generator does participate fully in shutdown, hangs shutdown


blt:
The `splunk_hec` generator spawns its own sub-tasks which stay active even while the rest of lading shut down. This leaves background tasks still running, which can hang the program on controlled shutdown. We do try to work around this by using a timed shutdown by this does not appear to work in all cases, as seen [here](https://github.com/vectordotdev/vector/runs/5665535762?check_suite_focus=true). The top-level process must signal to its spawned tasks that the time has come to shutdown.

blt:
It's possible that this is resolved by #203 but considering that we don't have tests in place to prove this I'll leave this ticket open.

PR #203, opened 2022-05-21
Resolve SplunkHec generator freeze bug

blt:
In https://github.com/vectordotdev/vector/pull/12776 we've seen that
the splunkhec generator doesn't quite work right with regard to
shutdown. This has been a long-running problem, see #155, and this
commit is intended to refactor the generator to be a little more
direct but, more importantly, have timeouts and definite loop breaks
in key places.

Whether this actually functions as expected will depend on Vector
where reproduction is the most feasible.


src/generator/splunk_hec/acknowledgements.rs
```diff
@@ -1,97 +1,120 @@
...
+                .await
+                .expect("receiver dropped unexpectedly"),
+        }
+    }
 }
 
 /// Splunk HEC channels
+#[derive(Debug)]
 pub(crate) struct Channels {
     /// If acknowledgements are enabled, the channel id maps to Some(Sender)
```
001wwang, line 52: Perhaps worth updating/moving this doc comment to reflect the new `Ack`/`NoAck` `Channel` setup.

src/generator/splunk_hec/acknowledgements.rs
```diff
@@ -103,85 +126,101 @@ impl AckService {
...
+                        let request: Request<Body> = Request::builder()
+                            .method(Method::POST)
+                            .uri(self.ack_uri.clone())
+                            .header(AUTHORIZATION, format!("Splunk {}", self.token))
+                            .header(SPLUNK_HEC_CHANNEL_HEADER, channel_id.clone())
+                            .body(body)
+                            .unwrap();
+                        let work = ack_request(self.client.clone(), request, channel_id.clone(), &mut ack_ids);
+
+                        if let Err(_err) = timeout(Duration::from_secs(1), work).await {
```
001wwang, line 167: Is it worth giving any extra timeout leeway here or is 1 second plenty?

001wwang (approved):
Thanks for refactoring this! Looks a lot tidier 👍

src/generator/splunk_hec/acknowledgements.rs
blt: Seemingly it is. At least, I was never able to trigger this in practice.

src/generator/splunk_hec/acknowledgements.rs
blt: Resolved in [16654f5](https://github.com/blt/lading/pull/203/commits/16654f5f7bb772c46173cf6a45faa7e0fc807bfa)

PR #205, opened 2022-05-24
Correct gauge collection

blt:
The gauge! macro from metrics requires an f64 but stores as an atomicu64. Our
collector failed to coerce this value properly, resulting in some hilariously
large gauges. I also realized that I could use serde's 'untagged' enum
representation to avoid casting floats to u64, saving some decimals if any
gauges are actually f64s.

### DIFF
--- a/src/captures.rs
+++ b/src/captures.rs
@@ -39,6 +39,16 @@ pub enum MetricKind {
     Gauge,
 }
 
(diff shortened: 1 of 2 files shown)


jszwedko (approved):
Looks good! Changes were easy to follow.

PR #220, opened 2022-06-07
[WIP] pattern generation, will require a lot of grinding

blt:


GeorgeHahn:
We should be able to do this with the [rand_regex](https://crates.io/crates/rand_regex) crate. This has been on my mind, I can swing around to it once the integration testing story is further along.

PR #230, opened 2022-06-27
Unify Payload configuration

blt:
Back when this project first started there were not many different kinds of
payload and only two generators. There was an awkward config -> payload switch
table that happened in each generator and, while I did love it, eh. There are
now more generators and more payloads with more on the way and what was once an
annoyance will quickly become a real code duplication problem.

This commit follows the generator/blackhole configuration model and has payload
provide its own configuration. This allows all the generators to drop their
Variant configuration and means introducing a new payload is not a big slog. If
we need generators to serialize differently from one another that is now more
difficult, but is surely something we can address by fiddling with the
Serializer trait.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -112,9 +112,9 @@ version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "10f203db73a71dfa2fb6dd22763990fa26f3d2625a6da2da900d23b87d26be27"
 dependencies = [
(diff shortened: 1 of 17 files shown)


GeorgeHahn (approved):
The code changes here LGTM, this looks like a timely improvement. I didn't test anything! I can take over this PR with testing & any fixes if you'd like - just assign it to me.

blt:
@GeorgeHahn all yours. :)

PR #248, opened 2022-07-28
Minor docs fixes

GeorgeHahn:
### What does this PR do?

Two tiny docs fixes

### Motivation

Really just testing out PR workflow & commit signing

### Related issues

N/A

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -57,10 +57,6 @@ pub enum ChunkError {
 ///
 /// Function will return an error if `block_byte_sizes` is empty or if a member
 /// of `block_byte_sizes` is large than `total_bytes`.
-///
-/// # Panics
-///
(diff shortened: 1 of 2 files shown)


src/block.rs
```diff
@@ -57,10 +57,6 @@ pub enum ChunkError {
 ///
 /// Function will return an error if `block_byte_sizes` is empty or if a member
 /// of `block_byte_sizes` is large than `total_bytes`.
-///
-/// # Panics
-///
-/// Function will panic if `block_bytes_sizes` is empty.
```
blt, line 63: Interesting. I'm honestly surprised that this doesn't cause a clippy ding, considering that there are a handful of `unwrap` instances present here.

blt (approved):

Issue #181, opened 2022-04-29
Target sub-process exit code should influence lading's


blt:
Right now if the target sub-process exits lading will log what the exit code was but will not actually fiddle with its own code. This means that lading can _appear_ to succeed at collecting samples while in practice this was not the case, in a way that harms use in CI.

PR #249, opened 2022-08-01
[SMP-21] Initial PID watching support

GeorgeHahn:
### What does this PR do?

This adds external process watching support via PID. The linux implementation uses a pidfd watcher to receive a notification and status code on exit. Other platforms use a polling approach that does not support status code retrieval.

### Motivation

This is the lading change required for container testing.

### Additional Notes

This is a **breaking change** for existing users. Binary targets now require a `--target-path` flag.


src/bin/lading.rs
```diff
@@ -71,12 +72,14 @@ struct Opts {
...
     global_labels: Option<CliKeyValues>,
+    /// measure an externally-launched process by PID
+    external_target_pid: Option<NonZeroU32>,
     /// additional environment variables to apply to the target, format
     /// KEY=VAL,KEY2=VAL
     #[clap(long)]
     target_environment_variables: Option<CliKeyValues>,
     /// the path of the target executable
-    target_path: PathBuf,
+    target_path: Option<PathBuf>,
```
blt, line 82: Might consider the use of an ArgGroup and to rely on clap's conflict logic: https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.conflicts_with

src/bin/lading.rs
```diff
@@ -286,8 +305,28 @@ async fn inner_main(
...
+            }
+            Target::PID(cfg) => {
+                let target_server = external_target::Server::new(cfg, shutdown.clone());
+                let tsrv = tokio::spawn(target_server.run(tgt_snd));
+                let res = tsrv.await;
+                if let Ok(Err(e)) = res {
+                    error!("pid target shut down with {:?}", e);
+                }
+            }
+        }
```
blt, line 328: These are pretty similar setups. I wonder if we couldn't maintain one `target` server and rely on that server to understand the difference between `Target::PID` and `Target::Binary`. I can see the two implementations slowly drifting?

GeorgeHahn: Agreed. I went back and forth on this a couple of times. There's just the right amount of code here to be in a grey area. I want to see what this looks like once it has the linux-specific pidfs code, but I like the idea of rolling them into one server.

Cargo.toml
```diff
@@ -44,6 +44,7 @@ uuid =  { version = "1.1", features = ["serde", "v4"] }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 procfs = { version = "0.12", default-features = false, features = [] }
+async-pidfd = "0.1"
```
blt, line 47: Neat crate.

src/target.rs
```diff
@@ -34,11 +41,23 @@ pub enum Error {
     Io(io::Error),
     /// Wrapper for [`nix::errno::Errno`]
     Errno(Errno),
+    /// The target PID does not exist or is invalid
+    PIDNotFound,
```
blt, line 45: Might be useful to include the pid in the error message, or wrap the original error message here.

src/target.rs
```diff
@@ -50,52 +69,166 @@ pub struct Config {
...
     ///
     /// None are known.
-    pub async fn run(mut self, pid_snd: Sender<u32>) -> Result<ExitStatus, Error> {
+    pub async fn run(self, pid_snd: Sender<u32>) -> Result<(), Error> {
         let config = self.config;
 
+        // Note that each target mode has different expectations around target
+        // exit. PID mode expects the target to continue running; any exit is
+        // a critical error. Binary mode expects the target to run until
+        // signalled, at which point it would prefer to see a graceful exit.
```
blt, line 138: 'it' being the mode?

blt (approved):
Some small comments / questions but this looks great.

src/target.rs
GeorgeHahn: Good idea. Added in 9df9c2a.

src/target.rs
GeorgeHahn: I deleted this sentence after reviewing it. Lading doesn't care what code the child exits with after receiving the signal and I don't see any reason to change that.

blt: There is https://github.com/DataDog/lading/issues/181 which might be relevant. Lading not reflecting the child exit code in some way does make CI failures harder to record.

GeorgeHahn: I think we should bubble up any exits that happen during the test. This is purely after the shutdown signal though - do we care what the child does after that point?

blt: Ah, I see the difference. Nah, I don't think so.

PR #261, opened 2022-08-08
Lading integration tests

GeorgeHahn:
### What does this PR do?

Adds an integration testing harness & target for lading. It's a lading sandwich!

`Sheepdog` guides the integration tests. `ducks` serves as a test target and follows along with `sheepdog's` instructions. Together, they're able to run `lading` in any given configuration and validate any interaction it has with a target.

### Motivation

This is the very start of an integration testing capability for `lading`. This should fairly quickly help us get to the point where `lading` changes can be validated with integration tests alone.

### Related issues

- [ ] Find/create Jira ticket


Cargo.toml
```diff
@@ -1,3 +1,11 @@
+[workspace]
+members = [
+    "./",
+    "integration/sheepdog",
```
blt, line 4: I am pro herding dogs.

integration/ducks/src/main.rs
```diff
@@ -0,0 +1,203 @@
...
+    let (_parts, body) = req.into_parts();
+    let body = body::to_bytes(body).await?;
+
+    {
+        let metric = GENERIC_HTTP_COUNTERS.get().unwrap().clone();
+        let mut m = metric.lock().unwrap();
+        m.request_count += 1;
+
+        m.total_bytes = body.len() as u64;
+        m.entropy.add(entropy::metric_entropy(&body) as f64);
```
GeorgeHahn, line 79: My thought here is that entropy may be a good enough indicator that `lading` is generating interesting data.

I'm currently seeing near-zero 50th percentile entropy measurements. I'm not sure if I'm holding something wrong, if I've misconfigured `lading`, or if the current configs are legitimately generating mostly low-entropy payloads. I'm going to do a quick validation by hand before really digging into this.

integration/shared/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+    None,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub enum AssertionConfig {
+    None,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DucksConfig {
```
GeorgeHahn, line 24: This isn't very deeply thought out and I think there's a very good chance we'll want to configure things differently before long. My gut feeling is this is easy enough to refactor that we can run with it and reshape it as we go.

integration/ducks/src/main.rs
blt: Considering that this has never been looked into beyond a quick "eh, looks good enough to me" some months ago it's entirely possible the payloads being generated are very low-entropy.

integration/ducks/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+entropy = "0.4"
+hyper = { version = "0.14", features = ["server"] }
+once_cell = "1"
+serde_json = "1"
+shared = { path = "../shared" }
+sketches-ddsketch = "0.1"
+tokio = { version = "1.20", features = ["rt", "macros", "fs", "io-util", "process", "signal", "time", "net"] }
+tokio-stream = { version = "0.1", features = ["net"] }
+tonic = "0.8"
+tower = { version = "0.4", default-features = false, features = ["timeout", "limit", "load-shed"] }
```
blt, line 27: As a result of working on Vector where the dependency tree is _huge_ I have a slight preference for this verbose sort of dependency configuration.

integration/ducks/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+aws-sdk-s3 = { version = "0.16", default-features = false, features = ["rustls", "rt-tokio"] }
+aws-sdk-sqs = { version = "0.16", default-features = false, features = ["rustls", "rt-tokio"] }
+aws-smithy-http = { version = "0.46", default-features = false, features = ["rt-tokio"] }
+aws-types = { version = "0.46", default-features = false, features = [] }
+
+anyhow = "1"
+entropy = "0.4"
+hyper = { version = "0.14", features = ["server"] }
+once_cell = "1"
+serde_json = "1"
```
blt, line 21: I'd like to specify the major/minor version of dependencies at least.

integration/ducks/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+
+static GENERIC_HTTP_COUNTERS: OnceCell<Arc<Mutex<GenericHttpCounters>>> = OnceCell::new();
+
+/// Corresponds to MeasurementConfig::GenericHttp
+struct GenericHttpCounters {
+    body_size: DDSketch,
+    entropy: DDSketch,
+    request_count: u64,
+    total_bytes: u64,
+    // record http methods received
```
blt, line 51: Intentionally left?

integration/ducks/src/main.rs
blt: The more I think on this the more I like the use of entropy as an indicator.

integration/ducks/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+        self.shutdown_tx.send(()).await.unwrap();
+        Ok(tonic::Response::new(Empty {}))
+    }
+
+    #[tracing::instrument(skip(self))]
+    async fn start_test(
+        &self,
+        _config: tonic::Request<TestConfig>,
+    ) -> Result<tonic::Response<shared::integration_api::ListenInfo>, Status> {
+        // todo get this from `config`
```
blt, line 173: Intentionally left?

integration/shared/src/lib.rs
blt: Reviewing this PR, that feeling does seem right to me, especially since we control every side of this arrangement.

integration/sheepdog/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+license = "MIT"
+publish = false
+
+[dependencies]
+anyhow = "1.0"
+assert_fs = "1.0.7"
+escargot = "0.5.7"
+serde_json = "1.0"
+shared = { path = "../shared" }
+thiserror = "1.0"
```
blt, line 16: `anyhow` and `thiserror`? I have a slight preference for `thiserror` but I would prefer to choose only one.

integration/sheepdog/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+        // only run this a single time though.
+        let ducks_binary = build_ducks()?;
+        let lading_binary = build_lading()?;
+
+        // Every ducks-sheepdog pair is connected by a unique socket file
+        let ducks_comm_file = self.tempdir.join("ducks_socket");
+
+        let mut ducks_process = Command::new(ducks_binary)
+            // switch Stdio to `inherit()` to see ducks logs in your terminal
+            .stdout(Stdio::piped())
```
blt, line 90: Might be valuable as a flag.

integration/sheepdog/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+        let channel = Endpoint::try_from("http://127.0.0.1/this-is-not-used")?
+            .connect_with_connector(tower::service_fn(move |_| {
+                UnixStream::connect(ducks_comm_file.clone())
+            }))
+            .await?;
+        let mut ducks_rpc = IntegrationTargetClient::new(channel);
+        debug!("connected to ducks");
+
+        // instruct ducks to start the test (this is currently hardcoded to a
+        // http sink test but will be configurable in the future)
```
blt, line 109: :+1:

blt (approved):
Very cool. I have some small nits inline but not blocking. Intention is to wire this all up into CI in a later PR, yeah?

integration/sheepdog/src/lib.rs
GeorgeHahn: I'm thinking about streaming these to sheepdog and printing them as they arrive. That will give them the same behavior as normal test output. That isn't urgent yet, but it'll be a big improvement once we have more integration tests in place.

blt: I like it.

integration/sheepdog/Cargo.toml
GeorgeHahn: Whoops, I wasn't even using thiserror. Same - I have a distinct preference for thiserror, but anyhow felt like a good fit here.

integration/ducks/src/main.rs
GeorgeHahn: Yeah, I wasn't too worried about this with the lack of implemented options right now. My next PR will resolve this.

integration/ducks/src/main.rs
GeorgeHahn: Sorry, I should have marked that todo. Yeah, I'll address this in a quick followup PR.

PR #273, opened 2022-08-17
TCP Integration Testing

GeorgeHahn:
### What does this PR do?

Add baseline support for TCP integration tests

### Motivation

This will allow more of the lading functionality to be tested

### Related issues

Followup to #261

### DIFF
--- a/integration/ducks/src/main.rs
+++ b/integration/ducks/src/main.rs
@@ -270,7 +270,7 @@ impl DucksTarget {
     }
 
     async fn tcp_listen(_config: DucksConfig, incoming: TcpListener) -> Result<(), anyhow::Error> {
-        debug!("TCP listener active");
+        debug!("TCP listener active on {}", incoming.local_addr()?);
         TCP_COUNTERS.get_or_init(|| Arc::new(Mutex::new(TcpCounters::default())));
 
         loop {
--- a/integration/ducks/src/main.rs
+++ b/integration/ducks/src/main.rs
@@ -263,8 +263,6 @@ impl DucksTarget {
                 m.read_count += 1;
                 m.total_bytes += buffer.len() as u64;
                 m.entropy.add(entropy::metric_entropy(&buffer) as f64);
-
-                // TODO: data rate metrics & a way to measure consistency
             }
 
             buffer.clear();
--- a/integration/ducks/src/main.rs
+++ b/integration/ducks/src/main.rs
@@ -248,7 +248,7 @@ impl DucksTarget {
     }
 
     async fn tcp_handler(mut socket: TcpStream) -> Result<(), anyhow::Error> {
-        // 500KB input buffer per connection (this can probably be smaller)
+        // 500KiB input buffer per connection (this can probably be smaller)
         let mut buffer = BytesMut::with_capacity(524_288);
         trace!("TCP connection received");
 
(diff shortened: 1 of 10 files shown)


integration/ducks/src/main.rs
```diff
@@ -113,107 +138,148 @@ async fn http_req_handler(req: Request<Body>) -> Result<hyper::Response<Body>, h
...
+            .timeout(Duration::from_secs(1))
+            .service(service);
+
+        let server = hyper::Server::builder(addr).serve(svc);
+        server.await?;
         Ok(())
     }
+
+    async fn tcp_handler(mut socket: TcpStream) -> Result<(), anyhow::Error> {
+        // 500KB input buffer per connection (this can probably be smaller)
```
blt, line 251: `KiB` properly, I think?

integration/ducks/src/main.rs
```diff
@@ -113,107 +138,148 @@ async fn http_req_handler(req: Request<Body>) -> Result<hyper::Response<Body>, h
...
+            socket.read_buf(&mut buffer).await?;
+
+            {
+                let metric = TCP_COUNTERS.get().unwrap();
+                let mut m = metric.lock().await;
+                m.read_count += 1;
+                m.total_bytes += buffer.len() as u64;
+                m.entropy.add(entropy::metric_entropy(&buffer) as f64);
+
+                // TODO: data rate metrics & a way to measure consistency
```
blt, line 267: Confirming this was intentionally left. Seems like.

integration/sheepdog/src/lib.rs
```diff
@@ -144,18 +162,32 @@ impl IntegrationTest {
...
+        let mut stdout_buf = String::new();
+        stdout.read_to_string(&mut stdout_buf).await?;
+        println!("ducks output:\n{}", stdout_buf);
+
+        let mut stdout = lading.stdout.unwrap();
+        let mut stdout_buf = String::new();
+        stdout.read_to_string(&mut stdout_buf).await?;
+        println!("lading output:\n{}", stdout_buf);
+
+        println!("test result: (todo)");
```
blt, line 186: Might be worth pulling `tracing` in at some point to handle output?

blt (approved):
Very nice. I had some comments, a question. I think when you're ready a bit in the README about how to add integration tests to the project is warranted.

integration/sheepdog/src/lib.rs
GeorgeHahn: I don't think we need `tracing` here yet, but we will before too long. These logs are emitted by `tracing` in the child, so they're already formatted nicely. This print gets them into cargo's test harness where they're only visible if the `--no-capture` flag is passed in.

This flow is going to change when we switch to RPC-streamed logs to support lading's binary-launch mode.

integration/ducks/src/main.rs
GeorgeHahn: It was, but I removed it after thinking about it some more. I need to read up on data rate calculations and may have some questions in our team channel. This isn't a terribly high priority at the moment and we might be able to get by with chunk size unit tests in lading.

GeorgeHahn:
I haven't been able to get a local repro for this failure; this even runs successfully in [act](https://github.com/nektos/act).

In Actions:
- Lading & ducks both start correctly
- Both run for the expected duration

I'm going to need to debug in-branch unfortunately. Sorry in advance for the noise.

blt:
> I'm going to need to debug in-branch unfortunately. Sorry in advance for the noise.

All good! Commits are free and I can always unsubscribe if it gets real noisy.

GeorgeHahn:
`lading` was trying to connect to ipv6 localhost 😭

blt (approved):

PR #278, opened 2022-08-17
Remove `tokio-process` & fix dependabot security advisories

GeorgeHahn:
### What does this PR do?

Remove the `tokio-process` dependency. Turns out it wasn't necessary - it's an old crate left over from when tokio used many separate crates.

### Motivation

This should close our dependabot security advisories.


blt (approved):
Neat.

PR #301, opened 2022-09-14
Support OTEL payloads

GeorgeHahn:
### What does this PR do?

Add baseline support for generating OTEL payloads. This can be used to generate binary OTLP/HTTP messages. OTLP/gRPC support will be built on top of this.

### Motivation

This is the first piece of OpenTelemetry support.

### Related issues

SMP-7

### Additional Notes

This takes on all of `tonic` as a dependency. That's not great - it's a heavy dependency and this PR only needs the codegen parts. I'm not too concerned about it because we're going to need tonic for `OTLP/gRPC` support. Related upstream issue: https://github.com/open-telemetry/opentelemetry-rust/issues/873.

### DIFF
--- a/src/payload/opentelemetry_log.rs
+++ b/src/payload/opentelemetry_log.rs
@@ -1,0 +1,181 @@
+//! Generates OpenTelemetry OTLP log payloads
+//!
+//! [Specification](https://opentelemetry.io/docs/reference/specification/protocol/otlp/),
+//! [data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md)
+//!
+//! This format is valid for OTLP/gRPC and binary OTLP/HTTP messages. The
+//! experimental JSON OTLP/HTTP format can also be supported but is not
+//! currently implemented.
+
+use crate::payload::{Error, Serialize};
+use arbitrary::{Arbitrary, Unstructured};
+use opentelemetry_proto::tonic::{
+    common::v1::{any_value, AnyValue},
+    logs::v1,
+};
+use prost::Message;
+use rand::Rng;
+use std::io::Write;
+
+/// Wrapper to generate arbitrary OpenTelemetry [`ExportLogsServiceRequests`](opentelemetry_proto::tonic::collector::logs::v1::ExportLogsServiceRequest)
+struct ExportLogsServiceRequest(Vec<LogRecord>);
+
+impl ExportLogsServiceRequest {
+    fn into_prost_type(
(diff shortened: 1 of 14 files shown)


blt (approved):
This looks excellent.

src/payload/opentelemetry_log.rs
```diff
@@ -0,0 +1,181 @@
...
+/// Wrapper to generate arbitrary OpenTelemetry [`ExportLogsServiceRequests`](opentelemetry_proto::tonic::collector::logs::v1::ExportLogsServiceRequest)
+struct ExportLogsServiceRequest(Vec<LogRecord>);
+
+impl ExportLogsServiceRequest {
+    fn into_prost_type(
+        self,
+    ) -> opentelemetry_proto::tonic::collector::logs::v1::ExportLogsServiceRequest {
+        opentelemetry_proto::tonic::collector::logs::v1::ExportLogsServiceRequest {
+            resource_logs: vec![v1::ResourceLogs {
+                resource: None,
```
spencergilbert, line 29: I can't imagine `None` vs some arbitrary set of k/v's here would make a huge performance difference, but it may be uncommon to not have any `resource` set.

src/payload/opentelemetry_metric.rs
```diff
@@ -0,0 +1,270 @@
...
+impl<'a> Arbitrary<'a> for Metric {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let name = u.arbitrary()?;
+        let description = u.arbitrary()?;
+        let unit = u.arbitrary()?;
+
+        let data = match u.int_in_range(0u8..=1)? {
+            0 => v1::metric::Data::Gauge(u.arbitrary::<Gauge>()?.0),
+            1 => v1::metric::Data::Sum(u.arbitrary::<Sum>()?.0),
+            // Currently unsupported: Histogram, ExponentialHistogram, Summary
```
spencergilbert, line 141: Is there a follow-up issue to support these three?

spencergilbert (approved):

src/payload/opentelemetry_metric.rs
GeorgeHahn: Good call, wrote up https://github.com/DataDog/lading/issues/303

src/payload/opentelemetry_log.rs
GeorgeHahn: 👍 #304

PR #327, opened 2022-09-20
Three fixes to the release workflow

GeorgeHahn:
### What does this PR do?

- I forgot to install the protobuf compiler
- The docker image is being tagged with the git tag verbatim, in addition to semver tags. This is unnecessary and confusing.
- <s>Stop tagging images with major version 0</s> Reverted. I can't delete the image that was released as '0', and I'm not too offended by having a '0' major tag. I can unwind this, but I'm not sure it's worth finding a github org admin to help fix.

### Motivation

Fixing up the release workflow

### Related issues

#320 #326 

### Additional Notes

<s>Unless anyone objects, I plan to delete the 0.10.0 tag, image, and release, retag, and allow the workflow to run again. I'm not against cutting a 0.10.1, but I'm not too concerned about re-tagging a dev version without code changes.</s> I'll cut a 0.10.1 release since I can't delete the 0.10.0 image.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -40,7 +40,7 @@ jobs:
             type=ref,prefix=pr-,event=pr
             type=semver,pattern={{version}},event=tag
             type=semver,pattern={{major}}.{{minor}},event=tag
-            type=semver,pattern={{major}},event=tag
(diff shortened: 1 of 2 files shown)


GeorgeHahn:
I might not be able to delete the docker image. That's reasonable but annoying because we definitely don't want a stale '0' tag hanging around forever.

blt (approved):

PR #329, opened 2022-09-20
WIP more release workflow fixes

GeorgeHahn:
### What does this PR do?

A brief description of the change being made with this pull request.

### Motivation

### Related issues

for this Pull Request.

### Additional Notes


.github/workflows/release.yml
```diff
@@ -24,28 +24,73 @@ jobs:
...
 
+      # Run the build & upload artifacts
       - name: Build and upload lading binaries
         uses: taiki-e/upload-rust-binary-action@v1
+
+        env:
+          CFLAGS: ${{ matrix.cflags }}
+          CXXFLAGS: ${{ matrix.cxxflags }}
+          CC: ${{ matrix.cc }}
+          CXX: ${{ matrix.cxx }}
```
GeorgeHahn, line 92: These should maybe be prefixed by `TARGET_`. It's not causing any issues today, but it could cause problems if we ever need to build C for the host from a build script. IIRC prost used to do that for the protobuf compiler and I had 'plenty' of 'fun' getting that build to work under yocto.

GeorgeHahn: Nevermind, `TARGET_CC`/`HOST_CC` and friends are from the cc crate, which the rdkafka crate does not use. We'll have some fun if we ever take on a C build time dependency.

blt (approved):

PR #342, opened 2022-10-02
UDP generator

GeorgeHahn:
### What does this PR do?

Implement a UDP generator & add some worst-case integration test timeouts.

### Motivation

This has been notably missing for a while, and it was a good project to work on from an airplane.

### Related issues

N/A

### Additional Notes

This does the absolute minimum for UDP payload size checking. That should be addressed more thoroughly when we add config validation.

### DIFF
--- a/src/generator/udp.rs
+++ b/src/generator/udp.rs
@@ -1,0 +1,197 @@
+//! The UDP protocol speaking generator.
+
+use std::{
+    net::{SocketAddr, ToSocketAddrs},
+    num::{NonZeroU32, NonZeroUsize},
+};
+
+use byte_unit::{Byte, ByteUnit};
+use governor::{
+    clock, state,
+    state::direct::{self, InsufficientCapacity},
+    Quota, RateLimiter,
+};
+use metrics::counter;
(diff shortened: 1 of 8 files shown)


src/generator/udp.rs
```diff
@@ -0,0 +1,197 @@
...
+        let labels = self.metric_labels;
+
+        let mut connection = None;
+        let mut blocks = self.block_cache.iter().cycle();
+
+        loop {
+            let blk = blocks.next().unwrap();
+            let total_bytes = blk.total_bytes;
+            assert!(
+                total_bytes.get() <= 65507,
```
blt, line 148: How'd we settle on 65,607? Maximum theoretical UDP packet size is 65,525 I think, less IPv4 header would be 65,509?

blt (approved):
One question. Not a major concern; real neat.

src/generator/udp.rs
GeorgeHahn: 0xffff - IPv4 header (minimum 20 B)- UDP header (8 B). It's unfortunate that this check is IPv4 specific.

Issue #101, opened 2021-10-15
Allow HTTP `blackhole` to respond with additional payloads. 


blt:
As of this writing `blackhole` has only a few response payloads:

* blank
* AWS Kinesis

It would be useful if the following responses were also available:

- [ ] Elasticsearch 
- [ ] S3
- [ ] Static byte sequence

Currently the body variants are here: https://github.com/blt/lading/blob/5acb3c419958536d43282ebfaa5c05042bd49b40/src/blackhole/http.rs#L32-L39

GeorgeHahn:
We may consider wiring this up to the new `payload` module if the implementation is reasonable. That will allow any payload to be used as a blackhole response. That would give us access to the generic payloads and offer a framework to build out specialized response payloads.

PR #358, opened 2022-10-20
Make http blackhole content type configurable

GeorgeHahn:
### What does this PR do?

Make the HTTP blackhole's response content type configurable.

### Motivation

This allows http responses to masquerade as anything. The next step will be to add configurable payloads, but we'll cross that bridge when we come to it.

### Related issues

https://github.com/DataDog/lading/issues/101

### Additional Notes

This required some unpleasantness to pass through to the service handler. I'd typically solve lifetimes for a series of moves with nested clones, but that's insufficient in this case. I worked around it by leaking a box and holding the `Copy` reference in the `FnMut` closure that was causing problems. I would gladly accept suggestions or improvements on this!

### DIFF
--- a/src/bin/lading.rs
+++ b/src/bin/lading.rs
@@ -274,7 +274,7 @@ async fn inner_main(
     //
     if let Some(cfgs) = config.blackhole {
         for cfg in cfgs {
-            let blackhole_server = blackhole::Server::new(cfg, shutdown.clone());
+            let blackhole_server = blackhole::Server::new(cfg, shutdown.clone()).unwrap();
             let _bsrv = tokio::spawn(async {
                 match blackhole_server.run().await {
                     Ok(()) => debug!("blackhole shut down successfully"),
--- a/src/blackhole/http.rs
+++ b/src/blackhole/http.rs
@@ -193,7 +193,13 @@ impl Http {
         // A very weird dance to satisfy the borrow checker for moving into & out of an FnMut.
         let content_type = self.content_type_header.clone();
         let content_type = Box::new(content_type);
-        // Turn the mutable ref from `leak` into an immutable ref so it's Copy
+        // Turn the mutable ref from `leak` into an immutable ref so it's Copy.
+        //
+        // This memory is intentionally leaked for the life of the process. This
+        // is acceptable because it only runs once at startup (for each
+        // configured `Http` generator). These generators are expected to stay
+        // active until the program shuts down. The memory is left for the OS
+        // to clean up.
         let content_type = &*Box::leak::<'static>(content_type);
 
         let service = make_service_fn(|_: &AddrStream| async move {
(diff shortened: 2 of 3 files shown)


GeorgeHahn:
Min repr of the content type unpleasantness: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3c23449db00a1d14ed834cdba7793404

src/bin/lading.rs
```diff
@@ -274,7 +274,7 @@ async fn inner_main(
     //
     if let Some(cfgs) = config.blackhole {
         for cfg in cfgs {
-            let blackhole_server = blackhole::Server::new(cfg, shutdown.clone());
+            let blackhole_server = blackhole::Server::new(cfg, shutdown.clone()).unwrap();
```
blt, line 277: If we don't want to kick out an error here might be nice to push an error message or something regarding the failure, or use `except`.

src/blackhole/http.rs
```diff
@@ -156,16 +183,23 @@ impl Http {
...
     ///
     /// # Panics
     ///
     /// None known.
     pub async fn run(mut self) -> Result<(), Error> {
+        // A very weird dance to satisfy the borrow checker for moving into & out of an FnMut.
+        let content_type = self.content_type_header.clone();
+        let content_type = Box::new(content_type);
+        // Turn the mutable ref from `leak` into an immutable ref so it's Copy
+        let content_type = &*Box::leak::<'static>(content_type);
```
blt, line 197: Spicy.

src/bin/lading.rs
GeorgeHahn: Pushed this to the backlog - other top level components also unwrap here, so I'd like to improve them all at once. https://github.com/DataDog/lading/issues/359

blt (approved):

src/blackhole/http.rs
```diff
@@ -190,23 +190,14 @@ impl Http {
...
-                srv(self.body_variant, request, content_type)
-            }))
+        let service = make_service_fn(|_: &AddrStream| {
+            let content_type = self.content_type_header.clone();
+            async move {
+                Ok::<_, hyper::Error>(service_fn(move |request| {
+                    debug!("REQUEST: {:?}", request);
+                    srv(self.body_variant, request, content_type.clone())
+                }))
+            }
```
GeorgeHahn, line 200: @blt This will clone on every new connection & on every response, versus on every successful response. I don't love that extra clone, but it's a pretty clear win for readability. I have a slight preference for this version unless you have perf concerns. What do you think?

blt: Unless we measure it as a performance problem I'm happy to go with something more readable.

blt: Much obliged, @bruceg.

PR #372, opened 2022-11-01
Bias UDP send loop towards the shutdown signal

GeorgeHahn:
### What does this PR do?

We've observed an issue where the UDP generator will run forever, ignoring the shutdown signal. This PR biases the UDP generation task towards polling the shutdown signal to avoid this.

### Motivation

Stop `lading` processes from spinning on UDP forever.

### Related issues

N/A

### Additional Notes

We may consider refactoring shutdown to abort async tasks rather than signal and expect cooperation.

### DIFF
--- a/src/generator/udp.rs
+++ b/src/generator/udp.rs
@@ -149,6 +149,7 @@ impl Udp {
                     "UDP packet too large (over 65507 B)"
                 );
                 if let Some(sock) = &connection {
+                    tokio::task::yield_now().await;
                     self.rate_limiter.until_n_ready(total_bytes).await.unwrap();
 
                     match sock.send_to(&blk.bytes, self.addr).await {
--- a/src/generator/udp.rs
+++ b/src/generator/udp.rs
@@ -150,6 +150,11 @@ impl Udp {
             );
 
             tokio::select! {
+                biased;
+                _ = self.shutdown.recv() => {
+                    info!("shutdown signal received");
+                    return Ok(());
+                },
                 sock = UdpSocket::bind("127.0.0.1:0"), if connection.is_none() => {
                     debug!("UDP port bound");
 
@@ -187,10 +192,6 @@ impl Udp {
                         }
                     }
                 }
-                _ = self.shutdown.recv() => {
-                    info!("shutdown signal received");
-                    return Ok(());
-                },
             }
         }
     }
--- a/src/generator/udp.rs
+++ b/src/generator/udp.rs
@@ -138,28 +138,36 @@ impl Udp {
         debug!("UDP generator running");
         let labels = self.metric_labels;
 
-        let mut connection = None;
-        let mut blocks = self.block_cache.iter().cycle();
+        let send_task = tokio::spawn(async move {
+            let mut blocks = self.block_cache.iter().cycle();
+            let mut connection = Option::<UdpSocket>::None;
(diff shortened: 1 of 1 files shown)


blt (approved):

GeorgeHahn:
This is pretty wild. Even rewritten as its own async task, the UDP sender causes other tasks on the runtime not to be polled. [a699c70](https://github.com/DataDog/lading/pull/372/commits/a699c708750096dd005c9c925add42221e3ad1a4) works around the issue by adding an explicit yield point. This is on my radar to min-repr and file upstream after further investigation.

---

To repro with lading:
- Set up a local lading target (I'm using an agent regression container)
- Add a log output to the UDP send loop
- Add an independent task that logs periodically
- Test with & without the yield point

Here's the lading config I'm using:

```yaml
generator:
  - udp:
      seed: [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53,
             59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131]
      addr: "127.0.0.1:10000"
      variant: "json"
      bytes_per_second: "500 Mb"
      block_sizes: ["1Kb"]
      maximum_prebuild_cache_size_bytes: "256 Mb"

blackhole:
  - http:
      binding_addr: "127.0.0.1:9091"
```

blt:
> Even rewritten as its own async task, the UDP sender causes other tasks on the runtime not to be polled. [a699c70](https://github.com/DataDog/lading/pull/372/commits/a699c708750096dd005c9c925add42221e3ad1a4) works around the issue by adding an explicit yield point. 

Dang. Getting upstream consideration of this would be excellent. I have wondered now and again if we might not profitably split lading's internals up so that each component runs on a single-threaded executor on top of an OS thread we manage from startup. We'd get better isolation, at least, for things like this. Non-trivial change, but it's crossed my mind now and again.

src/generator/udp.rs
```diff
@@ -149,6 +149,7 @@ impl Udp {
                     "UDP packet too large (over 65507 B)"
                 );
                 if let Some(sock) = &connection {
+                    tokio::task::yield_now().await;
```
blt, line 152: I guess at-speed we always have bytes ready, considering the size of the max payload. Hrm.

GeorgeHahn:
@blt sorry for the review re-request, I missed your comments earlier

blt (approved):

GeorgeHahn:
> Dang. Getting upstream consideration of this would be excellent. I have wondered now and again if we might not profitably split lading's internals up so that each component runs on a single-threaded executor on top of an OS thread we manage from startup. We'd get better isolation, at least, for things like this. Non-trivial change, but it's crossed my mind now and again.

This issue has me thinking about the same thing. Either a separate OS thread or another tokio runtime on its own pool of threads.

src/generator/udp.rs
GeorgeHahn: What's wild to me is that we have `cores - 1` tokio worker threads _sitting idle_ while this is happening. This should be very interesting to min-repr and dig into.

My initial thought is that we only have one task active. Maybe tokio tries to squeeze in its scheduling at a low priority after workers go idle. That could lead to exactly this situation.

PR #399, opened 2022-11-29
Tee target stdio to lading logs & release as 0.11.2

GeorgeHahn:
### What does this PR do?

This changes lading's child stdio forwarding to echo the target's logs in lading's logs as well as writing them to a file.

### Motivation

This will improve user visibility into target behavior.

### Related issues

SMP-242

### Additional Notes

I chose to implement this as an unconditional tee for the file output behavior because introducing new options here would require RJO changes. I suspect that this situation will evolve over the next few months to be more readily user-configurable, but at the moment adding new configs here doesn't seem meaningful and could introduce backwards compatibility issues.


blt (approved):
Tidy.

GeorgeHahn:
I'm going to put this on ice since we've solved the core issue using another strategy. This may still be desirable in the future, so I'll keep the branch around for now.

PR #421, opened 2022-12-20
Instrument `bytes_per_seconds`

blt:
### What does this PR do?

This PR instruments the bytes per seconds of each generator, save `file_tree`.

### Motivation

We would like to write out the maximum throughput that can be expected for a given
generator, where appropriate, into capture files to aid later analysis. Only
`file_tree` generator is not included here as its throughput concept is
completely alien to the network oriented generators.

### DIFF
--- a/src/generator/grpc.rs
+++ b/src/generator/grpc.rs
@@ -13,7 +13,7 @@ use governor::{
     Quota, RateLimiter,
 };
 use http::{uri::PathAndQuery, Uri};
-use metrics::counter;
+use metrics::{counter, gauge};
 use rand::rngs::StdRng;
 use rand::SeedableRng;
 use serde::Deserialize;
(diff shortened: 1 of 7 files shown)


src/generator/grpc.rs
```diff
@@ -195,9 +195,19 @@ impl Grpc {
...
             .map(|sz| NonZeroUsize::new(sz.get_bytes() as usize).expect("bytes must be non-zero"))
             .collect();
+        let labels = vec![
+            ("component".to_string(), "generator".to_string()),
+            ("component_name".to_string(), "grpc".to_string()),
+        ];
+
         let bytes_per_second = NonZeroU32::new(config.bytes_per_second.get_bytes() as u32).unwrap();
+        gauge!(
+            "Bytes_per_second",
```
GeorgeHahn, line 205: lowercase

src/generator/kafka.rs
```diff
@@ -164,6 +167,22 @@ impl Kafka {
             &labels,
         )?;
 
+        match config.throughput {
+            Throughput::BytesPerSecond { amount } => {
+                gauge!("bytes_per_second", amount.get_bytes() as f64, &labels);
+            }
+            Throughput::Unlimited => {
+                // The rate limiter we takes a NonZeroU32. Our 'unlimited'
```
goxberry, line 175: The rate limiter we use?

GeorgeHahn (approved):
This looks great

src/generator/grpc.rs
blt: Addressed in [32ba992](https://github.com/DataDog/lading/pull/421/commits/32ba9920a85edc4caff92b49f082a89e69abb373)

src/generator/kafka.rs
blt: We use a library called https://docs.rs/governor/latest/governor/ for most of our generators to limit the activity the generators do. At startup a generator uses its throughput setting to pre-generate all its load in 'block' sizes and then round-robins through those blocks, blocking on the governor until there's enough quota to move forward with the size of the current block.

GeorgeHahn: You're missing the word `use` though 😜

GeorgeHahn: @blt I canceled auto-merge so you have a minute to fix this if desired

blt: Haha I a verb.

PR #424, opened 2022-12-27
introduce process tree generator

safchain:
### What does this PR do?

This PR adds a process tree generator. The goal is to generate a process tree with variable width and height. The leaves execute a command line specified in the config file with static args/envs or generated ones. A typical "pstree like" gives : 

```
1994835  437791  |   |   |   |       \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
1999492 1994835  |   |   |   |           \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
1999545 1999492  |   |   |   |               \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
1999914 1999545  |   |   |   |               |   \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
1999547 1999492  |   |   |   |               \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
1999915 1999547  |   |   |   |               |   \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
1999610 1999492  |   |   |   |               \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
1999917 1999610  |   |   |   |               |   \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
1999850 1999492  |   |   |   |               \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
1999916 1999850  |   |   |   |               |   \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
1999908 1999492  |   |   |   |               \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
1999918 1999908  |   |   |   |               |   \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
1999919 1999492  |   |   |   |               \_ ./target/debug/lading --config-path ./examples/lading.yaml --target-pid 83568
``` 

In a follow-up PR cgroups handling will be introduce.

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### DIFF
--- a/src/bin/lading.rs
+++ b/src/bin/lading.rs
@@ -7,16 +7,21 @@ use std::{
     str::FromStr,
 };
 
-use clap::{ArgGroup, Parser};
+use clap::{ArgGroup, Parser, Subcommand};
 use lading::{
     blackhole,
     captures::CaptureManager,
     config::{Config, Telemetry},
-    generator, inspector, observer,
+    generator::{
+        self,
+        process_tree::{self},
+    },
+    inspector, observer,
     signals::Shutdown,
     target::{self, Behavior, Output},
 };
 use metrics_exporter_prometheus::PrometheusBuilder;
+use rand::{rngs::StdRng, SeedableRng};
 use tokio::{
     runtime::Builder,
     signal,
@@ -123,10 +128,33 @@ struct Opts {
     /// whether to ignore inspector configuration, if present, and not run the inspector
     #[clap(long)]
     disable_inspector: bool,
+    /// Extra sub commands
+    #[clap(subcommand)]
+    extracmds: Option<ExtraCommands>,
 }
 
-fn get_config() -> (Opts, Config) {
-    let ops: Opts = Opts::parse();
+#[derive(Subcommand, Debug)]
+#[clap(hide = true)]
+enum ExtraCommands {
+    ProcessTreeGen(ProcessTreeGen),
+}
+
+#[derive(Parser, Debug)]
+#[clap(group(
+    ArgGroup::new("config")
+        .required(true)
+        .args(&["config-path", "config-content"]),
+))]
+struct ProcessTreeGen {
+    /// path on disk to the configuration file
+    #[clap(long)]
+    config_path: Option<PathBuf>,
+    /// string repesanting the configuration
+    #[clap(long)]
+    config_content: Option<String>,
+}
+
+fn get_config(ops: &Opts) -> Config {
     debug!(
         "Attempting to open configuration file at: {}",
         ops.config_path
@@ -192,7 +220,7 @@ fn get_config() -> (Opts, Config) {
             }
         }
     }
-    (ops, config)
+    config
 }
 
 async fn inner_main(
@@ -334,11 +362,61 @@ async fn inner_main(
     shutdown.wait(max_shutdown_delay).await;
 }
 
+fn run_process_tree(opts: ProcessTreeGen) {
+    let mut contents = String::new();
+
+    if let Some(path) = opts.config_path {
+        debug!(
+            "Attempting to open configuration file at: {}",
+            path.display()
+        );
+        let mut file: std::fs::File = std::fs::OpenOptions::new()
+            .read(true)
+            .open(&path)
+            .unwrap_or_else(|_| panic!("Could not open configuration file at: {}", path.display()));
+
+        file.read_to_string(&mut contents).unwrap();
+    } else if let Some(str) = &opts.config_content {
+        contents = str.to_string()
+    } else {
+        unreachable!("clap ensures that exactly one target option is selected");
+    };
+
+    match process_tree::get_config(&contents) {
+        Ok(config) => {
+            info!("Generating a process tree.");
+
+            let mut rng = StdRng::from_seed(config.seed);
+            let nodes = process_tree::generate_tree(&mut rng, &config);
+
+            process_tree::spawn_tree(&nodes, config.process_sleep_ns.get());
+
+            info!("Bye. :)");
+        }
+        Err(e) => panic!("invalide configuration: {}", e),
+    }
+}
+
+fn run_extra_cmds(cmds: ExtraCommands) {
+    match cmds {
+        ExtraCommands::ProcessTreeGen(opts) => run_process_tree(opts),
+    }
+}
+
 fn main() {
     tracing_subscriber::fmt::init();
 
     info!("Starting lading run.");
-    let (opts, config): (Opts, Config) = get_config();
+    let opts: Opts = Opts::parse();
+
+    // handle extra commands
+    if let Some(cmds) = opts.extracmds {
+        run_extra_cmds(cmds);
+        return;
+    }
+
+    let config = get_config(&opts);
+
     let experiment_duration = Duration::from_secs(opts.experiment_duration_seconds.into());
     let warmup_duration = Duration::from_secs(opts.warmup_duration_seconds.into());
     // The maximum shutdown delay is shared between `inner_main` and this
--- a/src/bin/lading.rs
+++ b/src/bin/lading.rs
@@ -399,6 +399,7 @@ fn run_process_tree(opts: ProcessTreeGen) {
 
 fn run_extra_cmds(cmds: ExtraCommands) {
     match cmds {
+        // This command will call fork and the process must be kept fork-safe up to this point.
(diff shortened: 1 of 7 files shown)


src/generator/process_tree.rs
```diff
@@ -0,0 +1,383 @@
...
+                    info!("shutdown signal received");
+                    break;
+                },
+            }
+        }
+        Ok(())
+    }
+}
+
+fn spawn_tree(nodes: &VecDeque<Process>, sleep_ns: u32) {
```
blt, line 281: One immediate problem here is that this function needs to be `async`, else the loop beginning on 266 doesn't properly function and the whole of lading might not either. This function is thread blocking and doesn't time share properly, meaning if lading is run on a single core other tasks in lading will be denied access to the CPU, even if `thread::sleep` is the only thing happening. 

`try_wait_pid` uses the nix crate to wait on pids, although I think this generator could be written in terms of [`tokio::process`](https://docs.rs/tokio/latest/tokio/process/) and avoid the blocking nature. If there's some nuance I'm missing, possibly around cgroup addition, then calling [`spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) with `spawn_tree` would be the right move.

src/generator/process_tree.rs
```diff
@@ -0,0 +1,383 @@
...
+
+fn default_envs_count() -> NonZeroU32 {
+    NonZeroU32::new(10).unwrap()
+}
+
+#[derive(Debug, Deserialize, PartialEq, Eq, Clone)]
+/// Configuration of [`ProcessTree`]
+pub struct Args {
+    /// The mode, can be static or generate
+    pub mode: String,
```
blt, line 80: Please convert this to an enum. You'd have in this module:

```rust
enum Mode {
  Static,
  Generate
}
```

and then below when you do string comparisons you can instead check for the enum variant, much cheaper and type-safe.

src/generator/process_tree.rs
```diff
@@ -0,0 +1,383 @@
...
+        if process.is_exec() {
+            let status = Command::new(&process.executable)
+                .args(&process.args)
+                .envs(&process.envs)
+                .stdout(Stdio::null())
+                .stderr(Stdio::null())
+                .status()
+                .ok()
+                .unwrap();
+            exit(status.code().unwrap())
```
blt, line 315: This `exit` will kill the whole of lading. Other sinks have a careful shut-down process and we exit only for catastrophic errors. It's not clear to me why we need to exit with the child's status here?

src/generator/process_tree.rs
```diff
@@ -0,0 +1,383 @@
...
+            executable: exec.executable.to_string(),
+            args,
+            envs,
+            depth,
+        }
+    }
+
+    fn is_exec(&self) -> bool {
+        !self.executable.is_empty()
+    }
```
blt, line 222: Considering that this is called not infrequently it would be better to add an `is_exec` boolean in the struct. Also, under what conditions does a `Process` exist but not have an executable? @GeorgeHahn is better with clap than me but if the concern is config validation we can ensure the user can't set an empty string here.

blt (changes requested):
I have some comments but this is good stuff. If some of the comments about blocking don't make sense I'm happy to pair up.

GeorgeHahn (changes requested):
Thanks for the PR! This is going in a good direction. My main concern is around the `fork` at the core of the tree generator. Extracting that should also address the comments Brian raised around async & child shutdown.

src/generator/process_tree.rs
safchain: Thanks @blt I'll convert it to an async version.

src/generator/process_tree.rs
safchain: Since we are in the is_exec case it means we are in the child(leaf) process. In this child process we exec(fork/exec) and then we exit. I don't think it exit the whole lading process(main). At least this is not what I observed when I developed it but I could have miss something.

safchain:
@blt , @GeorgeHahn thanks for the review, that is really useful. I'll work on addressing your comments.

src/generator/process_tree.rs
safchain: Good point. I'll add a boolean and add a check to ensure that the path is an executable

src/generator/process_tree.rs
blt: On further examination I was confused and I believe you are correct.

safchain:
@blt @GeorgeHahn I updated the PR. I tried to address your comments. Now the process generator is made of a hidden  subcommand.

src/generator/process_tree.rs
```diff
@@ -0,0 +1,524 @@
...
+            tokio::select! {
+                _ = process_rate_limiter.until_ready() => {
+                    // using pid as target pid just to pass laging clap constraints
+                    let _res = Command::new(lading_path)
+                        .args(["--target-pid", "1"])
+                        .arg("process-tree-gen")
+                        .arg("--config-content")
+                        .arg(&self.config_content)
+                        .stdin(Stdio::null())
+                        .spawn();
```
GeorgeHahn, line 312: This should emit a log message or return an error if spawn fails

GeorgeHahn (commented):
This is looking great!

src/generator/process_tree.rs
```diff
@@ -0,0 +1,524 @@
...
+    pub max_tree_per_second: NonZeroU32,
+    /// The maximum depth of the process tree
+    #[serde(default = "default_max_depth")]
+    pub max_depth: NonZeroU32,
+    /// The maximum children per level
+    #[serde(default = "default_max_children")]
+    pub max_children: NonZeroU32,
+    /// Sleep applied at process start
+    #[serde(default = "default_process_sleep_ns")]
+    pub process_sleep_ns: NonZeroU32,
```
blt, line 227: This could be a [`Duration`](https://doc.rust-lang.org/std/time/struct.Duration.html) directly. Not strictly necessary, more of a comment.

src/generator/process_tree.rs
blt: +1, I think an error so we can shutdown quickly.

src/generator/process_tree.rs
```diff
@@ -0,0 +1,383 @@
...
+                .envs(&process.envs)
+                .stdout(Stdio::null())
+                .stderr(Stdio::null())
+                .status()
+                .ok()
+                .unwrap();
+            exit(status.code().unwrap())
+        }
+
+        match unsafe { fork() } {
```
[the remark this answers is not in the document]
blt: @GeorgeHahn I think this is resolved?

blt (approved):
This looks great to me. I had one small comment but I'd be glad to see this go in as-is.

src/bin/lading.rs
```diff
@@ -334,11 +362,61 @@ async fn inner_main(
...
+
+            info!("Bye. :)");
+        }
+        Err(e) => panic!("invalide configuration: {}", e),
+    }
+}
+
+fn run_extra_cmds(cmds: ExtraCommands) {
+    match cmds {
+        ExtraCommands::ProcessTreeGen(opts) => run_process_tree(opts),
```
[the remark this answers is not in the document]
safchain: Done

src/generator/process_tree.rs
safchain: I would like to do it but it will make the configuration parsing a bit more complex to parse the Duration

src/generator/process_tree.rs
safchain: Done

src/generator/process_tree.rs
```diff
@@ -0,0 +1,556 @@
...
+        let lading_path = self.lading_path.to_str().unwrap();
+
+        loop {
+            tokio::select! {
+                _ = process_rate_limiter.until_ready() => {
+                    // using pid as target pid just to pass laging clap constraints
+                    let output = Command::new(lading_path)
+                        .args(["--target-pid", "1"])
+                        .arg("process-tree-gen")
+                        .arg("--config-contdqdsent")
```
GeorgeHahn, line 334: Accidental addition here

safchain: oups, thanks, fixed

GeorgeHahn (approved):

PR #427, opened 2023-01-04
Remove kafka generator support

blt:
### What does this PR do?

This commit removes kafka generator support from lading. We believe it it unused and its maintenance burden is non-trivial, especially with consideration to the kafka crate's unstable interface.

### Additional Notes

This is a break changing and so lading's version is now bumped to 0.12.


jszwedko (approved):
👍 It doesn't seem like Vector is using it.

I could see testing Vector's Kafka performance being important to us in the future but we can revisit how to implement it when we get to that point. I could see it being preferable to actually use Kafka as part of testing. There are other sources which similarly integrate with external systems like: `aws_s3` and `aws_sqs` that might be better tested with those systems. Just food for thought.

GeorgeHahn (changes requested):
The release workflow needs some adjustment for this change: https://github.com/DataDog/lading/blob/main/.github/workflows/release.yml#L31-L33

blt:
> The release workflow needs some adjustment for this change: https://github.com/DataDog/lading/blob/main/.github/workflows/release.yml#L31-L33

Good catch.

blt:
> > The release workflow needs some adjustment for this change: https://github.com/DataDog/lading/blob/main/.github/workflows/release.yml#L31-L33
> 
> Good catch.

Addressed in [6d9ce2c](https://github.com/DataDog/lading/pull/427/commits/6d9ce2c6c7d7060262fe41b6c52d619720fee4ce).

GeorgeHahn (approved):

PR #432, opened 2023-01-07
Introduce DogStatsD payload generation

blt:
This commit introduces DogStatsD payload generation into lading. We support all the variants described [here](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics). 

### DIFF
--- a/src/payload/dogstatsd/common.rs
+++ b/src/payload/dogstatsd/common.rs
@@ -1,0 +1,154 @@
+use std::{collections::HashMap, mem};
+
+use arbitrary::Unstructured;
+
+const MAX_SMALLVEC: usize = 8;
+const MAX_TAGS: usize = 16;
+const SIZES: [usize; 8] = [1, 2, 4, 8, 16, 32, 64, 128];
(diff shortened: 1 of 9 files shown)


src/payload/dogstatsd/common.rs
```diff
@@ -0,0 +1,178 @@
...
+impl fmt::Display for MetricTagStr {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
+}
+
+impl<'a> arbitrary::Arbitrary<'a> for MetricTagStr {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let choice: u8 = u.arbitrary()?;
+        let size = SIZES[(choice as usize) % SIZES.len()];
```
GeorgeHahn, line 38: Minor concern that this would give unevenly distributed choices if `SIZES` changes length in the future. How about `u.int_in_range(0..SIZES.len())`?

GeorgeHahn (approved):

src/payload/dogstatsd/common.rs
blt: Much better.

PR #439, opened 2023-01-11
Make Unix Datagram, Stream distinct, fix DogStatsD bugs

blt:
### What does this PR do?

This commit improves over the UDS generator, splitting it firmly between stream and datagram implementations. I have confirmed that this works now against Datadog Agent, requiring some fixes to the DogStatsD payload generation.


blt:
I got a little greedy here. I can split the payload changes into a separate PR if that's desirable.

GeorgeHahn (approved):

PR #448, opened 2023-01-30
Update file generator to use `select!` approach

blt:
Peer to #447 this commit adjusts the file generator to use the `select!` approach, now allowing for clean shutdown of the file generator sub-tasks with consistent flushing. This commit also properly allows errors to bubble up from those sub-tasks, where previously they would not.

REF SMP-329

### DIFF
--- a/src/generator/file_gen.rs
+++ b/src/generator/file_gen.rs
@@ -15,6 +15,7 @@ use std::{
 };
 
 use byte_unit::{Byte, ByteUnit};
+use futures::future::join_all;
 use governor::{
     clock, state,
     state::direct::{self, InsufficientCapacity},
@@ -46,7 +47,7 @@ pub enum Error {
     Io(::std::io::Error),
     /// Creation of payload blocks failed.
     Block(block::Error),
-    /// Could not join on child task results.
+    /// Child sub-task error.
     Child(JoinError),
 }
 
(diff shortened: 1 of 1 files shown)


src/generator/file_gen.rs
```diff
@@ -214,7 +223,7 @@ impl FileGen {
         self.shutdown.recv().await;
         info!("shutdown signal received");
         for handle in self.handles.drain(..) {
-            handle.abort();
+            handle.await??;
```
GeorgeHahn, line 226: I don't think this is terribly important, but this shutdown would be more graceful if it waited for all tasks to complete before returning errors.

GeorgeHahn (approved):

src/generator/file_gen.rs
blt: Ah yeah, that's true. I suppose it's not more work to add a `join_all` here, although I did think when I originally wrote it that seeing _an_ error message concerning a sub-task crash was probably more explicable than seeing all? But I might be wrong there.

GeorgeHahn: `join_all` and returning the first error seems pretty reasonable here. That's usually what I end up doing by way of the `Vec<Result<T, E>>` -> `Result<Vec<T>, E>` `collect` implementation.

blt: Adjusted in [6a60cf2](https://github.com/DataDog/lading/pull/448/commits/6a60cf2ff3c9b63d842c42716cf2acce566508a0)

PR #466, opened 2023-02-09
Capacity threshold search

blt:
### What does this PR do?

This pull request replaces our throttle mechanism based on [governor](https://docs.rs/governor/latest/governor/) with a similar cell rate algorithm but one where the 'capacity' of the leaky bucket shifts in response to actual use. This allows lading to automatically search out stable throughput points in clients. 

### Motivation

We're experimenting on Agent and discovered that in some configurations -- UDS with DogStatsD, notably -- we could force the Agent's ingress throughput to demonstrate high standard deviation. At low throughput -- 256 KiB/sec -- this problem went away, implying there was stable point somewhere between 1 GiB/sec and 256 KiB/sec. We want lading to search this point out automatically.

### Related issues

REF SMP-372

### DIFF
--- a/integration/sheepdog/src/lib.rs
+++ b/integration/sheepdog/src/lib.rs
@@ -257,7 +257,11 @@ generator:
 
         let reqs = test.run().await?;
 
-        assert!(reqs.http.request_count > 10);
+        debug_assert!(
+            reqs.http.request_count > 10,
+            "Request count: {request_count}",
+            request_count = reqs.http.request_count
+        );
         Ok(())
     }
 
@@ -289,7 +293,11 @@ generator:
 
         let reqs = test.run().await?;
 
-        assert!(reqs.http.request_count > 10);
+        debug_assert!(
+            reqs.http.request_count > 10,
+            "Request count: {request_count}",
+            request_count = reqs.http.request_count
+        );
         Ok(())
     }
 
@@ -321,7 +329,11 @@ generator:
 
         let reqs = test.run().await?;
 
-        assert!(reqs.http.request_count > 10);
+        debug_assert!(
+            reqs.http.request_count > 10,
+            "Request count: {request_count}",
+            request_count = reqs.http.request_count
+        );
         Ok(())
     }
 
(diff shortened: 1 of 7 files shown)


src/throttle.rs
```diff
@@ -19,46 +18,398 @@ pub(crate) enum Error {
...
+
+        // Now that the preliminaries are out of the way, wake up and compute
+        // how much the throttle capacity is refilled since we were last
+        // called. Depending on how long ago this was we may have completely
+        // filled up throttle capacity. Note we fill to the _projected_ budget
+        // and not the maximum capacity.
+        let ticks_since = self.clock.ticks_elapsed();
+        let refilled_capacity: u64 = cmp::min(
+            ticks_since
+                .wrapping_mul(f64::MAX as u64)
```
GeorgeHahn, line 209: Could you explain what this line is doing?

src/throttle.rs
```diff
@@ -19,46 +18,398 @@ pub(crate) enum Error {
...
+                .wrapping_mul(self.refill_per_tick)
+                .wrapping_add(self.spare_capacity),
```
GeorgeHahn, line 211: Should these be saturating operations? Saturation seems preferable to potentially wrapping to a value smaller than `self.projected_budget`

blt: Ah, you're absolutely right!

src/throttle.rs
blt: This is maybe the funniest bug I've ever written, is what this is doing.

blt:
CC @scottopell @ddyurchenko

integration/sheepdog/src/lib.rs
```diff
@@ -227,7 +228,11 @@ generator:
 
         let reqs = test.run().await?;
 
-        assert!(reqs.http.request_count > 10);
+        debug_assert!(
+            reqs.http.request_count > 10,
+            "Request count: {request_count}",
+            request_count = reqs.http.request_count
+        );
```
GeorgeHahn, line 235: Is this change desirable? It looks like it would nullify the test if it is ever run in release mode. (same for other `assert` -> `debug_assert` changes.)

GeorgeHahn (approved):

integration/sheepdog/src/lib.rs
blt: Ah, you are right. No, this is not a desirable change. I got confused here with the usual test macros not disabling.

PR #493, opened 2023-03-08
Add Proportional Set Size (PSS) support to observer

GeorgeHahn:
### What does this PR do?

Aggregate the PSS value for all maps in all processes under observation. This will give a more accurate view of overall memory usage for process hierarchies that share memory.

### Motivation

Improve our memory measurements


blt (approved):
Very cool.

GeorgeHahn:
I'm concerned about the ubuntu integration test failure here. This works fine for me locally. I'm going to dig into this tomorrow.

PR #499, opened 2023-03-14
Remove prometheus export of metrics

blt:
### What does this PR do?

The use of the prometheus exporter was a convenience early on in this project but has not seen use for at least six months by our known user-base. 

### Motivation

Removing this resolves a ding we have for
[RUSTSEC-2020-0168](https://rustsec.org/advisories/RUSTSEC-2020-0168) and slightly simplifies our CLI flags.


goxberry (approved):

blt:
Ah, we still get dinged even so. Closing.

PR #506, opened 2023-03-21
Introduce support for TraceAgent payload

blt:
### What does this PR do?

This commit introduces support for the Datadog Agent tracing span, defined [here](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/pb/span.proto). In practice this will be combined with the HTTP generator to POST payloads toward the Agent.

### Additional Notes

As of this commit -- as noted in the code -- the shrink algorithm in our payload creation isn't quite right, but I'm sure something will come to me tomorrow.

### Related issues

REF SMP-410

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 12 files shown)


blt:
@ahmed-mez FYI

GeorgeHahn (approved):

PR #527, opened 2023-04-03
Run regression detector with the release candidate

GeorgeHahn:
### What does this PR do?

Update 'outer' lading with support for experiments that don't have a generator.

### Motivation

This allows `apache_common_http_to_blackhole` to function.

### Related issues

#514


GeorgeHahn:
I'm going to split the prepare-release part of this out to a separate PR as soon as the RC binaries are published. That will allow us to hold 0.13.2 until the RC is tested in the trusted CI flow.

PR #528, opened 2023-04-03
Prepare release 0.13.2

GeorgeHahn:
### What does this PR do?

Release 0.13.2 that was tested in #527 

---

## Added
- Added a CLI flag to disable the target module
- Allow lading to run without any generators active


GeorgeHahn:
Leaving this unreleased. The existing RC is usable for the changes that would have been released here and `main` has moved past this minor version.

PR #532, opened 2023-04-04
Allow for customizable payloads

blt:
### What does this PR do?

This commit starts converting our payload generation away from the `arbitrary` crate. While it's served us well for a long time it lacks in a few areas:

* customization by the user of data in-payload,
* ability for generation to reference 'up the tree'

These two points are especially useful for payloads like dogstatsd where we would want to specify the number of metric names and the total number of unique tag combinations.

### Related issues

SMP-463


src/payload/common.rs
```diff
@@ -1,15 +1,58 @@
...
+use super::Generator;
 
 const SIZES: [usize; 12] = [1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048];
 const CHARSET: &[u8] = b"abcdefghijklmnopqrstuvwxyz ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789().,";
 #[allow(clippy::cast_possible_truncation)]
 const CHARSET_LEN: u8 = CHARSET.len() as u8;
 
+impl DistString for AsciiStr {
+    fn append_string<R: Rng + ?Sized>(&self, rng: &mut R, s: &mut String, len: usize) {
+        s.reserve(4 * len); // max size of an `AsciiChar`
```
GeorgeHahn, line 13: Looks like a stale comment. `AsciiChar` -> `char`?

GeorgeHahn (commented):
This is looking good

src/payload/common.rs
blt: Good eye, yes.

src/payload/dogstatsd.rs
```diff
@@ -1,29 +1,100 @@
...
+            small_strings: random_strings_with_length(1024, 8, &mut rng),
+        }
+    }
+}
+
+impl<'a> Generator<Member<'a>> for MemberGenerator {
+    fn generate<R>(&'a self, rng: &mut R) -> Member<'a>
+    where
+        R: rand::Rng + ?Sized,
+    {
```
blt, line 69: @GeorgeHahn I've gotten stuck on a lifetime error here. I wonder if you could take a look? My hope is to convert this payload to where we calculate all the metric names etc at the top in `MemberGenerator` and then pass those down through the sub-generators without needing to reallocate etc. So far so good but I can't get the lifetime of `Member<'a>` and `&'b self` to line up to where `a == b`.

GeorgeHahn: I'm not sure this is possible with how the `Generator` trait is currently setup.

```rust
pub(crate) trait Generator<I> {
    // Note that the `self` reference has an implicit lifetime
    fn generate<'a, R>(&self, rng: &mut R) -> I 
    where
        R: rand::Rng + ?Sized;
}
```

```rust
impl<'a> Generator<Event<'a>> for EventGenerator<'a> {
    // lifetime mismatch - function returns an `Event<'1>`. Fixing this requires an explicit lifetime on `self` and possibly also a lifetime on the return value.
    fn generate<R>(&self, mut rng: &mut R) -> Event<'a> 
    where
        R: rand::Rng + ?Sized,
    {
        let title = self.titles.choose(&mut rng).unwrap();
        let text = self.texts_or_messages.choose(&mut rng).unwrap();

        Event {
            title_utf8_length: title.len(),
            text_utf8_length: text.len(),
            title,
            text,
...
```

There's also an issue of overlapping lifetime names that adds confusion. The `generate` trait function defines a lifetime `'a` and its implementations define another lifetime with the same name. Bumping that lifetime up to the trait level should help align the lifetimes and get more meaningful error messages.

blt: Hah! I scooted the lifetime down from the trait to get a clearer error message. If we could pair on this today briefly I think that would be a big help to me.

blt: After we chatted I ended up taking the approach we agreed on: lots of little clones are okay in this situation.

GeorgeHahn (approved):

PR #573, opened 2023-04-28
Introduce configurable throttle

blt:
### What does this PR do?

This commit allows generators to provide a configuration option for their throttle, here only udp is converted. The ultimate ambition is to allow for two additional throttle variants: all-out and stable. We rename the existing throttle to 'predictive' and set this as the default. Note that the throttle is implemented in terms of an enum. I did attempt to make this static dispatch because throttle is called in our hot loops but found the intrusion into the codebase too high, especially with consideration of the throttle being present determined at runtime.

### Related issues

REF SMP-520
REF SMP-521

### DIFF
--- a/src/throttle.rs
+++ b/src/throttle.rs
@@ -9,6 +9,7 @@ mod predictive;
 
 #[derive(Debug, Deserialize, PartialEq, Clone, Copy)]
 /// Configuration of this generator.
+#[serde(rename_all = "snake_case")]
 pub enum Config {
     /// Create a throttle that predicts target capability
     Predictive,
--- a/src/throttle.rs
+++ b/src/throttle.rs
@@ -1,21 +1,30 @@
 //! A throttle for input into the target program.
 
 use async_trait::async_trait;
-use metrics::gauge;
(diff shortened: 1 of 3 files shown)


GeorgeHahn (approved):

src/throttle.rs
```diff
@@ -1,21 +1,30 @@
...
 
-use crate::target;
+mod predictive;
 
-const INTERVAL_TICKS: u64 = 1_000_000;
+#[derive(Debug, Deserialize, PartialEq, Clone, Copy)]
+/// Configuration of this generator.
+pub enum Config {
+    /// Create a throttle that predicts target capability
+    Predictive,
```
GeorgeHahn, line 14: This enum should have a snake case serde rename rule

blt: Right you are.

blt: Resolved in [ee97732](https://github.com/DataDog/lading/pull/573/commits/ee97732a8b86af9b90cc6f34c9d5ac0f41094b67)

PR #592, opened 2023-05-24
Component IDs

GeorgeHahn:
### What does this PR do?

Add optional component IDs to the top-level generator & blackhole config.

### Motivation

This will allow us to differentiate between components.

### Related issues

SMP-575

### DIFF
--- a/src/blackhole/http.rs
+++ b/src/blackhole/http.rs
@@ -1,6 +1,6 @@
 //! The HTTP protocol speaking blackhole.
 
-use std::{net::SocketAddr, time::Duration};
+use std::{net::SocketAddr, sync::Arc, time::Duration};
 
 use http::{header::InvalidHeaderValue, status::InvalidStatusCode, HeaderMap};
 use hyper::{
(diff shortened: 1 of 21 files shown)


src/blackhole/http.rs
```diff
@@ -114,11 +104,12 @@ struct KinesisPutRecordBatchResponse {
 #[allow(clippy::borrow_interior_mutable_const)]
 async fn srv(
     status: StatusCode,
+    labels: Arc<Vec<(String, String)>>,
     body_variant: BodyVariant,
     req: Request<Body>,
     headers: HeaderMap,
 ) -> Result<Response<Body>, hyper::Error> {
-    requests_received().increment(1);
+    counter!("bytes_received", 1, &*labels);
```
blt, line 112: Hmm, I hate to go back to `counter!`. It's actually not terribly cheap to create these things on the fly like this. At least I managed to drop lading's CPU use by ~30% when generating dogstatsd metrics over UDS.

blt: I wonder, could we maybe avoid passing `labels` and instead pass the metrics handle itself, pre-loaded with labels?

blt (commented):
On the whole this looks excellent. I do regret the need to go back to the use of `counter!` and `gauge!` and wonder if we couldn't contrive to avoid them.

src/blackhole/http.rs
GeorgeHahn: Good call, that worked nicely.

blt (approved):

PR #613, opened 2023-07-05
Slightly tweak how observer performs math operations

blt:
### What does this PR do?

This commit slightly tweaks how we tally up our tick and number of thread data. Instead of immediately converting to a u64 we sum every interior i64 and only then take the unsigned absolute value. The code is a little more tidy than it was and at the extreme end of the value range we avoid losing some information.

### Related issues

REF SMP-612

### DIFF
--- a/src/observer.rs
+++ b/src/observer.rs
@@ -161,8 +161,8 @@ impl Server {
                         let uptime_seconds: f64 = Uptime::new().expect("could not query uptime").uptime;
                         let process_uptime_seconds = uptime_seconds - process_starttime_seconds;
 
-                        let cutime: u64 = all_stats.iter().map(|stat| <i64 as std::convert::TryInto<u64>>::try_into(stat.0.cutime).unwrap()).sum();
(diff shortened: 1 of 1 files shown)


goxberry (approved):

goxberry:
Makes sense to me. IIRC, the typical pitfall of signed arithmetic would be wrapping due to overflow, which would likely pollute results worse than unsigned overflow, but that case seems unlikely here, given that `i64::MAX` is around 9e18. Even assuming 100 threads, overflow would require more than 25 trillion hours per thread, which seems...unlikely.

blt:
> Makes sense to me. IIRC, the typical pitfall of signed arithmetic would be wrapping due to overflow, which would likely pollute results worse than unsigned overflow, but that case seems unlikely here, given that `i64::MAX` is around 9e18. Even assuming 100 threads, overflow would require more than 25 trillion hours per thread, which seems...unlikely.

Agreed. I do find it undesirable but, yeah, really unlikely.

PR #616, opened 2023-07-05
Update observer to collect CPU utilization data

blt:
### What does this PR do?

This commit updates the observer with the aim of calculating CPU utilization. We maintain output of user and kernel space CPU seconds but add utilization with the understanding that this is useful to users that need to consider utilization and not ticks.

### Additional Notes

Note, if there is CPU use in the system that is not in a parent, child relationship but should still be counted this change will not address that.

### Related issues

REF SMP-613
#613

### DIFF
--- a/src/observer.rs
+++ b/src/observer.rs
@@ -145,7 +145,7 @@ impl Server {
             .map_err(Error::ProcError)?;
 
         let limits = process.limits().map_err(Error::ProcError)?;
-        // NOTE units on the CPU limits are 'CPU / second'
+        // NOTE units on the CPU limits are 'CPU-seconds / second'
         let max_cpu_time: Limit = limits.max_cpu_time;
         let soft_cpu_limit: f64 = match max_cpu_time.soft_limit {
             LimitValue::Unlimited => f64::MAX,
@@ -188,18 +188,18 @@ impl Server {
                         let stime: u64 = all_stats.iter().map(|stat| stat.0.stime).sum();
 
                         let kernel_time_ticks: u64 = cstime.unsigned_abs() + stime; // CPU-ticks
-                        let kernel_time_seconds: f64 = kernel_time_ticks as f64 / ticks_per_second as f64; // CPU
+                        let kernel_time_seconds: f64 = kernel_time_ticks as f64 / ticks_per_second as f64; // CPU-seconds
                         let user_time_ticks: u64 = cutime.unsigned_abs() + utime; // CPU-ticks
-                        let user_time_seconds: f64 = user_time_ticks as f64 / ticks_per_second as f64; // CPU
+                        let user_time_seconds: f64 = user_time_ticks as f64 / ticks_per_second as f64; // CPU-seconds
 
                         let process_uptime_seconds_diff: f64 = process_uptime_seconds - prev_process_uptime_seconds; // second
                         let kernel_time_ticks_diff = (kernel_time_ticks - prev_kernel_time_ticks) as f64; // CPU-ticks
                         let user_time_ticks_diff = (user_time_ticks - prev_user_time_ticks) as f64; // CPU-ticks
                         let time_ticks_diff = (kernel_time_ticks + user_time_ticks) - (prev_kernel_time_ticks + prev_user_time_ticks); // CPU-ticks
 
-                        let kernel_time_seconds_diff: f64 = kernel_time_ticks_diff / ticks_per_second as f64; // CPU / second
-                        let user_time_seconds_diff: f64 = user_time_ticks_diff / ticks_per_second as f64; // CPU / second
-                        let time_seconds_diff: f64 = time_ticks_diff as f64 / ticks_per_second as f64; // CPU / second
+                        let kernel_time_seconds_diff: f64 = kernel_time_ticks_diff / ticks_per_second as f64; // CPU-seconds
+                        let user_time_seconds_diff: f64 = user_time_ticks_diff / ticks_per_second as f64; // CPU-seconds
+                        let time_seconds_diff: f64 = time_ticks_diff as f64 / ticks_per_second as f64; // CPU-seconds
 
                         let kernel_utilization_soft = (kernel_time_seconds_diff / process_uptime_seconds_diff) / soft_cpu_limit;
                         let kernel_utilization_hard = (kernel_time_seconds_diff / process_uptime_seconds_diff) / hard_cpu_limit;
--- a/src/observer.rs
+++ b/src/observer.rs
@@ -133,10 +133,7 @@ impl Server {
         use std::{sync::atomic::Ordering, time::Duration};
 
         use metrics::{gauge, register_counter, register_gauge};
-        use procfs::{
-            process::{Limit, LimitValue},
-            Uptime,
-        };
+        use procfs::Uptime;
 
         let target_pid = pid_snd
             .recv()
--- a/src/observer.rs
+++ b/src/observer.rs
@@ -55,6 +55,27 @@ pub struct Server {
     shutdown: Shutdown,
 }
 
+#[inline]
+#[cfg(target_os = "linux")]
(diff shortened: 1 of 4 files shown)


src/observer.rs
```diff
@@ -156,25 +175,55 @@ impl Server {
...
+                        // The utilization of CPU time in kernel-space with regard to hard cgroup CPU/second limit
+                        gauge!("kernel_time_utilization_hard", kernel_utilization_hard);
+                        // The utilization of CPU time in user-space with regard to soft cgroup CPU/second limit
+                        gauge!("user_time_utilization_soft", user_utilization_soft);
+                        // The utilization of CPU time in user-space with regard to hard cgroup CPU/second limit
+                        gauge!("user_time_utilization_hard", user_utilization_hard);
+                        // The utilization of CPU time in user-space and kernel-space with regard to soft cgroup CPU/second limit
+                        gauge!("cpu_time_utilization_soft", utilization_soft);
+                        // The utilization of CPU time in user-space and kernel-space with regard to hard cgroup CPU/second limit
+                        gauge!("cpu_time_utilization_hard", utilization_hard);
```
goxberry, line 222: Are these values effectively unitless? (By that, I mean they'd be percentages if we multiplied by 100.)

src/observer.rs
```diff
@@ -156,25 +175,55 @@ impl Server {
...
 
-                        let kernel_time_seconds = (cstime + stime) as f64 / ticks_per_second;
-                        let user_time_seconds = (cutime + utime) as f64 / ticks_per_second;
+                        let kernel_time_seconds: u64 = (cstime.unsigned_abs() + stime)  / ticks_per_second; // CPU
+                        let user_time_seconds: u64 = (cutime.unsigned_abs() + utime)  / ticks_per_second; // CPU
+
+                        let process_uptime_seconds_diff: f64 = process_uptime_seconds - prev_process_uptime_seconds; // second
+                        let kernel_time_seconds_diff = (kernel_time_seconds - prev_kernel_time_seconds) as f64; // CPU
+                        let user_time_seconds_diff = (user_time_seconds - prev_user_time_seconds) as f64; // CPU
+                        let time_seconds_diff = ((kernel_time_seconds + user_time_seconds) - (prev_kernel_time_seconds + prev_user_time_seconds)) as f64; // CPU
```
goxberry, line 196: If I'm understanding the computations correctly -- and I may not be -- it seems like (most of) the raw timing data is in ticks, which have integer values? If so, it seems like the best precision that could possibly be achieved in representing the same value in seconds is `(1.0f64 / (procfs::ticks_per_second() as f64))`, which is `0.01f64`, as far as I understand it. My concern here would be subtractive cancellation polluting the precision of each of the expressions of the form `x - y`, where `x` and `y` both have type `f64` (and in this case, `x` and `y` would both have units of seconds).

A more faithful calculation of these values would be to defer the conversion of ticks to seconds to as late as possible within the loop because integer subtraction is exact (assuming no overflow). In other words, I'd compute the differences in ticks (because the integer subtraction would be exact), then convert them to seconds, instead of the current approach that converts ticks to seconds before subtraction.

src/observer.rs
blt: They are.

src/observer.rs
blt: You know, that is a good point. I'll adjust.

src/observer.rs
```diff
@@ -141,13 +144,29 @@ impl Server {
         let process = Process::new(target_pid.try_into().expect("PID coercion failed"))
             .map_err(Error::ProcError)?;
 
-        let ticks_per_second: f64 = procfs::ticks_per_second() as f64;
+        let limits = process.limits().map_err(Error::ProcError)?;
+        // NOTE units on the CPU limits are 'CPU / second'
+        let max_cpu_time: Limit = limits.max_cpu_time;
```
goxberry, line 149: Does this value come from the `RLIMIT_CPU` field of the `struct rlimit` queried by the `getrlimit` syscall ([man page](https://man7.org/linux/man-pages/man2/getrlimit.2.html))? I'm wondering if the units of this value are seconds or CPU*seconds.

src/observer.rs
```diff
@@ -156,25 +175,55 @@ impl Server {
...
+                        let kernel_time_seconds_diff = (kernel_time_seconds - prev_kernel_time_seconds) as f64; // CPU
+                        let user_time_seconds_diff = (user_time_seconds - prev_user_time_seconds) as f64; // CPU
+                        let time_seconds_diff = ((kernel_time_seconds + user_time_seconds) - (prev_kernel_time_seconds + prev_user_time_seconds)) as f64; // CPU
+
+                        let kernel_utilization_soft = (kernel_time_seconds_diff / process_uptime_seconds_diff) / soft_cpu_limit;
+                        let kernel_utilization_hard = (kernel_time_seconds_diff / process_uptime_seconds_diff) / hard_cpu_limit;
+                        let user_utilization_soft = (user_time_seconds_diff / process_uptime_seconds_diff) / soft_cpu_limit;
+                        let user_utilization_hard = (user_time_seconds_diff / process_uptime_seconds_diff) / hard_cpu_limit;
+                        let utilization_soft = (time_seconds_diff / process_uptime_seconds_diff) / soft_cpu_limit;
+                        let utilization_hard = (time_seconds_diff / process_uptime_seconds_diff) / hard_cpu_limit;
```
goxberry, line 203: Related to [my comment above asking about the units of `max_cpu_time` above](https://github.com/DataDog/lading/pull/616/files#r1253608560), I'm wondering if the division by `process_uptime_seconds_diff` is necessary in this group of calculations.

src/observer.rs
blt: Adjusted in [b4723bc](https://github.com/DataDog/lading/pull/616/commits/b4723bcd9a88a101f41ca95bd130bbe4dec1cf81)

src/observer.rs
blt: It's read out of `/proc/pid/limits`, see https://docs.rs/procfs/0.15.1/src/procfs/process/limit.rs.html. So yeah, `RLIMIT_CPU`. "This is a limit, in seconds, on the amount of CPU time that the process can consume." Which I think would be, in the notation I'm using here, CPU units per second.

goxberry (approved):

src/observer.rs
```diff
@@ -156,25 +175,61 @@ impl Server {
...
+                        let kernel_time_seconds: f64 = kernel_time_ticks as f64 / ticks_per_second as f64; // CPU
+                        let user_time_ticks: u64 = cutime.unsigned_abs() + utime; // CPU-ticks
+                        let user_time_seconds: f64 = user_time_ticks as f64 / ticks_per_second as f64; // CPU
+
+                        let process_uptime_seconds_diff: f64 = process_uptime_seconds - prev_process_uptime_seconds; // second
+                        let kernel_time_ticks_diff = (kernel_time_ticks - prev_kernel_time_ticks) as f64; // CPU-ticks
+                        let user_time_ticks_diff = (user_time_ticks - prev_user_time_ticks) as f64; // CPU-ticks
+                        let time_ticks_diff = (kernel_time_ticks + user_time_ticks) - (prev_kernel_time_ticks + prev_user_time_ticks); // CPU-ticks
+
+                        let kernel_time_seconds_diff: f64 = kernel_time_ticks_diff / ticks_per_second as f64; // CPU / second
```
goxberry, line 200: I agree with all of the units comments in the loop up to this one. I get `CPU-seconds` because `ticks_per_second` should have units of `ticks / seconds`.

blt: Agree.

blt: Adjusted the comment units to specify specifically `CPU-seconds` and correct this mistake in [ac005df](https://github.com/DataDog/lading/pull/616/commits/ac005df68666beeea024468e395cb2ee5fd2ec7b).

goxberry (approved):

PR #629, opened 2023-07-12
Split project into sub-crates, add capture protobuf

blt:
### What does this PR do?

This commit splits the project into two sub-crates, one 'lading' which is familiar and one 'lading-capture' which has the traditional JSON form of a `Line` now exported and a new protobuf based capture payload intended for network transmission.

### Related issues

REF SMP-631


regression/Dockerfile
```diff
@@ -4,6 +4,7 @@
 FROM docker.io/rust:1.68.1-bullseye@sha256:ce1f9cc5f88955a7384a957a64793012945a7c130883edab0af44168723c7b77 as builder
 WORKDIR /work
 COPY . .
+RUN apt-get update && apt-get install -y protobuf-compiler && rm -rf /var/lib/apt/lists/*
```
GeorgeHahn, line 7: This needs to be added in the release workflow as well. It's surprisingly notrivial to get right for mac & musl builds.

blt: Ah, right you are.

.github/workflows/release.yml
```diff
@@ -59,6 +53,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Install protobuf (Apt)
+        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
+        if: matrix.os == 'ubuntu-20.04'
+
+      - name: Install protobuf (Brew)
+        run: brew install protobuf
+        if: matrix.os == 'macos-latest'
```
GeorgeHahn, line 62: No action needed, but something to note for memory's sake: if the `protoc` setup steps get annoying, we could change things around so we don't need it available in CI. The specific strategy varies; the essence is that the commits will include the rust changes corresponding to any proto changes that were made. We could flag `protoc` out of the CI build and use the generated rust files as-is. We'd probably want to add a hash of the proto files to the rust files to ensure they line up. The [opentelemetry-proto](https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-proto) rust crate uses a similar strategy.

Cargo.toml
```diff
@@ -1,99 +1,16 @@
 [workspace]
-members = ["./", "integration/sheepdog", "integration/ducks"]
+members = [
+        "./",
```
GeorgeHahn, line 3: Is `./` necessary to include in the workspace?

lading_capture/proto/capture/v1/capture.proto
```diff
@@ -0,0 +1,38 @@
...
+  double value = 4;
+  // Labels associated with this line.
+  map<string, string> labels = 5;
+}
+
+// A collection of Lines
+message Payload {
+  // An ID that is unique to a single lading run.
+  string run_id = 1;
+  // The time in milliseconds that this Payload was collected.
```
GeorgeHahn, line 32: `milliseconds past the epoch`?

lading_capture/proto/capture/v1/capture.proto
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// A capture value container, called a 'line' for historical reasons.
+message Line {
+  // The name of the metric recorded by this line.
+  string name = 2;
+  // The kind of metric recorded by this line.
+  MetricKind kind = 3;
+  // The value of the metric recorded by this line.
+  double value = 4;
```
GeorgeHahn, line 23: The JSON `Line` can specify a floating point or integer value (corresponding to a gauge or a counter, respectively). I'm open to keeping this as-is, but I thought I'd call it out. What do you think?

GeorgeHahn (approved):
This looks great! I left a couple of minor comments, nothing blocking.

lading_capture/proto/capture/v1/capture.proto
blt: I originally put that there out of an abundance of caution but considering that we _already_ do wacky float / integer conversions in and out of metrics-rs I don't think were were actually gaining anything.

Cargo.toml
blt: It's a good question. Let me fiddle with it.

.github/workflows/release.yml
blt: That's not a bad idea.

PR #636, opened 2023-07-18
Allow lading config to be specified via env var

scottopell:
### What does this PR do?
Supports an environment variable `LADING_CONFIG` that can contain the exact same yaml as the current config. The environment variable takes precedence over the on-disk config.

### Motivation
This is useful in a kubernetes deployment file to avoid the need for a config-map to put a file on disk. It keeps all config tightly coupled to the `lading` pod.

### Related issues
none.

### Additional Notes

This is useful to me, but I have no strong opinions if its a good fit for lading upstream. I defer to your vision for lading whether or not you want to merge this.


blt (approved):

blt:
@scottopell I like it. Would you add this into the CHANGELOG?

PR #639, opened 2023-07-20
Collect procfs data per-process

blt:
### What does this PR do?

This commit modifies how we collect procfs data: we now distinguish children from their parent. Previous versions of lading rolled this information up into one time series but were now distinguish them with labels. This allows users interested only in the roll-up to continue sum over names if they wish.

### Additional Notes

We have dropped tick series. It saves some room in the capture file and tick information was not found to be useful in practice.

### Related issues

REF SMP-631

### DIFF
--- a/lading/src/observer/linux.rs
+++ b/lading/src/observer/linux.rs
@@ -252,14 +252,12 @@ fn percentage(delta_ticks: f64, delta_time: f64, num_cores: f64) -> f64 {
     }
 
     let mut overall_percentage = (delta_ticks / delta_time) * 100.0;
-    if overall_percentage > 100.0 {
-        overall_percentage = 100.0;
+    let percent = overall_percentage / num_cores;
+    if percent > 100.0 {
+        100.0
+    } else if percent < 0.0 {
+        0.0
+    } else {
+        percent
     }
-
-    let mut percent = overall_percentage * num_cores;
-    if percent < 0.0 {
-        percent = 0.0;
-    }
-
-    percent
 }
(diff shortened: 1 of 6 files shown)


lading/src/observer/linux.rs
```diff
@@ -0,0 +1,241 @@
...
+            // further by trying to account for the behavior of the operating
+            // system. Anyway, good luck out there. You'll be fine.
+            let rss: u64 = stats.rss;
+            let rsslim: u64 = stats.rsslim;
+            let vsize: u64 = stats.vsize;
+
+            let labels = [("pid", format!("{pid}")), ("exe", basename)];
+
+            // Number of pages that the process has in real memory.
+            gauge!("rss_bytes", rss as f64, &labels);
```
blt, line 174: Oops, this needs to be multiplied by page_size.

lading/src/observer/linux.rs
```diff
@@ -0,0 +1,243 @@
...
+    if delta_time == 0.0 {
+        return 0.0;
+    }
+
+    let mut overall_percentage = (delta_ticks / delta_time) * 100.0;
+    if overall_percentage > 100.0 {
+        overall_percentage = 100.0;
+    }
+
+    let mut percent = overall_percentage * num_cores;
```
GeorgeHahn, line 259: I don't follow this part of the calculation. The justification in `agent` is `Calculates utilization split across all CPUs. A busy-loop process on a 2-CPU-core system would be reported as 50% instead of 100%.`. Why? It seems to me that it'd be reported as 100% in the code here. (That also explains the 800% number that we're seeing when this saturates.)

GeorgeHahn (approved):

PR #647, opened 2023-07-21
Allow users to configure metric multivalue weights

blt:
### What does this PR do?

In DogStatsD it is allowable for a metric line to have multiple values. In practice, this is rare. Our previous 1/32 odds of producing only a single-valued line makes lading's output unusual. This is not a bad thing but it is behavior that we want users to be able to configure.

For the purposes of backward compatibility we maintain the same 1/32 odds.

### Related issues

REF SMP-663

### DIFF
--- a/lading/src/payload/dogstatsd.rs
+++ b/lading/src/payload/dogstatsd.rs
@@ -127,7 +127,7 @@ pub struct Config {
     /// Defines the relative probability of each kind of DogStatsD metric.
     #[serde(default)]
     pub metric_weights: MetricWeights,
-    /// Defines the relative probability of a metric having multiple values.
+    /// Defines the relative probability of a dogstatsd message having multiple values.
     /// Choices are weighted according to the specified weight.
     #[serde(default = "default_metric_multivalue")]
     pub metric_multivalue: Vec<MetricValueWeight>,
--- a/lading/src/payload/dogstatsd.rs
+++ b/lading/src/payload/dogstatsd.rs
@@ -35,6 +35,13 @@ fn default_tag_keys_maximum() -> usize {
     64
 }
 
+fn default_metric_multivalue() -> Vec<u8> {
+    vec![
+        1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25,
+        26, 27, 28, 29, 30, 31,
+    ]
+}
+
 /// Weights for `DogStatsD` kinds: metrics, events, service checks
 ///
 /// Defines the relative probability of each kind of `DogStatsD` datagram.
@@ -82,7 +89,7 @@ impl Default for MetricWeights {
 }
 
 /// Configure the `DogStatsD` payload.
-#[derive(Debug, Deserialize, Clone, Copy, PartialEq)]
+#[derive(Debug, Deserialize, Clone, PartialEq)]
 pub struct Config {
     /// Defines the minimum number of metric names allowed in a payload.
     #[serde(default = "default_metric_names_minimum")]
@@ -105,6 +112,9 @@ pub struct Config {
     /// Defines the relative probability of each kind of DogStatsD metic.
     #[serde(default)]
     pub metric_weights: MetricWeights,
+    /// Defines the relative probability of a metric having multiple values.
+    #[serde(default = "default_metric_multivalue")]
+    pub metric_multivalue: Vec<u8>,
 }
 
 fn choose_or_not<R, T>(mut rng: &mut R, pool: &[T]) -> Option<T>
@@ -145,6 +155,7 @@ impl MemberGenerator {
         key_range: Range<usize>,
         kind_weights: KindWeights,
         metric_weights: MetricWeights,
+        metric_multivalue: Vec<u8>,
         mut rng: &mut R,
     ) -> Self
     where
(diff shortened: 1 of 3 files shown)


lading/src/payload/dogstatsd.rs
```diff
@@ -105,6 +112,9 @@ pub struct Config {
     /// Defines the relative probability of each kind of DogStatsD metic.
     #[serde(default)]
     pub metric_weights: MetricWeights,
+    /// Defines the relative probability of a metric having multiple values.
+    #[serde(default = "default_metric_multivalue")]
+    pub metric_multivalue: Vec<u8>,
```
GeorgeHahn, line 117: This doc comment and the implementation don't quite match up for me. This is tough to describe, maybe something like 'multi-value metrics are evenly weighted among sizes specified in this list.'

GeorgeHahn (approved):

lading/src/payload/dogstatsd.rs
goxberry: Agreed.

blt: I like that.

goxberry: I'd also say the current specification could make it slightly awkward to specify some discrete probability distributions, but it depends on how much flexibility you want to give users. If I wanted to have a 75% chance of selecting a single-valued metric and a 25% chance of selecting a metric with 2 values, I'd have to input `vec![1, 1, 1, 2]`, or some permutation of that, if I'm understanding this code correctly.

goxberry: Put another way, it seems like each entry in the vector has an equal probability of selection, but each distinct value in the vector does *not* have an equal probability of selection, if there are duplicates. "Evenly weighted among sizes specified in this list" seems ambiguous as to how duplicate values are treated.

blt: I'm okay with that. The alternative would be to make a struct that has a value and its respective weight, which would also work fine. @scottopell is our primary user here, curious what your thoughts are.

blt: Hmm actually no this is unduly cumbersome.

blt:
```
generator:
  - udp:
      seed: [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53,
             59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131]
      addr: "127.0.0.1:8125"
      throttle: stable
      variant:
        dogstatsd:
          metric_names_minimum: 32
          metric_names_maximum: 128
          tag_keys_minimum: 0
          tag_keys_maximum: 512
          kind_weights:
            metric: 90
            event: 5
            service_check: 5
          # NOTE these ratios are calculated from our live data
          # `datadog.dogstatsd.client.metrics_by_type` and should be considered
          # a company secret. They are representative only of our private,
          # company usage and not of our customers'.
          metric_weights:
            count: 100
            gauge: 10
            timer: 0
            distribution: 50
            set: 0
            histogram: 1
          metric_multivalue:
            - { value: 1, weight: 20 }
            - { value: 2, weight: 2 }
            - { value: 4, weight: 1 }
      bytes_per_second: "100 Mb"
      block_sizes: ["256b", "512b", "1Kb", "1432b"]
      maximum_prebuild_cache_size_bytes: "500 Mb"

blackhole:
  - http:
      binding_addr: "127.0.0.1:9091"
  - http:
      binding_addr: "127.0.0.1:9092"

target_metrics:
  - expvar:
      uri: "http://127.0.0.1:5000/debug/vars"
      vars:
        - "/dogstatsd-udp/Bytes"
        - "/dogstatsd-udp/Packets"
        - "/dogstatsd-udp/PacketReadingErrors"
```

From [9aa4b6b](https://github.com/DataDog/lading/pull/647/commits/9aa4b6b6a3375bb445119b7dd9bc71574c891451)

goxberry: If the assumption is that the number of metric values is small, would something like a vector of nonnegative integers work? Maybe something like `weight: Vec<u8>`, such that the probability of generating a metric with `i` values is `(&weight[i] as f64) / weight.iter().sum::<f64>()`? (A vector of nonnegative floating-point numbers would also work.)

I had something like `rand_distr::weighted_alias::WeightedAliasIndex` in mind, but I could also see taking a `HashSet` of a vector of lengths, and then selecting one of the elements of the `HashSet` with equal probability, if that restriction wouldn't be unduly burdensome to users.

blt: I think that could also work. I do think it makes figuring out what the probability of a pick is gets tough quick, so I have a preference for what we have in 9aa4b6b just because I find it easier to reason about at a glance. If there's mood for `WeightedAliasIndex` or similar I could see that too.

scottopell:
This looks like it'll do what we need. I'm curious what the best default would be, but 1..32 works for now.

I may request/make a change in the future to allow specifying this per-metric type, but no need for that at the moment. Note that `SET` dogstatsd messages don't support multi-value, I created https://github.com/DataDog/lading/issues/650 about this.

lading/src/payload/dogstatsd.rs
```diff
@@ -105,6 +127,10 @@ pub struct Config {
     /// Defines the relative probability of each kind of DogStatsD metic.
     #[serde(default)]
     pub metric_weights: MetricWeights,
+    /// Defines the relative probability of a metric having multiple values.
+    /// Choices are evenly weighted among those listed.
+    #[serde(default = "default_metric_multivalue")]
+    pub metric_multivalue: Vec<MetricValueWeight>,
```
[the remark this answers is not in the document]
blt: That's the correct understanding, yes.

PR #659, opened 2023-07-31
Update lading to 0.17.3

blt:


GeorgeHahn (approved):
I believe this is missing a cargo.lock change. Otherwise LGTM.

PR #660, opened 2023-08-01
Introduce property tests for stable throttle

blt:
### What does this PR do?

We have, lately, wondered if the stable throttle was giving too much capacity to its clients. This PR is an attempt to answer that question through property testing. Note I have separated out the async and non-async bits of `Stable` and am property testing the non-async bit, making it feasible to do this kind of examination. I suspect I can do the same for the predictive throttle.

REF SMP-664


blt:
@scottopell FYI

lading/src/throttle/stable.rs
```diff
@@ -57,75 +46,224 @@ where
...
+    fn request(&mut self, ticks_elapsed: u64, capacity_request: NonZeroU32) -> Result<u64, Error> {
+        let capacity_request: u64 = u64::from(capacity_request.get());
         // Okay, here's the idea. At the base of `Stable` is a cell rate
         // algorithm. We have bucket that gradually fills up and when it's full
         // it doesn't fill up anymore. Callers draw down on this capacity and if
         // they draw down more than is available in the bucket they're made to
         // wait.
 
-        gauge!(
-            "throttle_refills_per_tick",
```
GeorgeHahn, line 67: There's a doc note for this metric that should be deleted from the top of the file.

lading/src/throttle/stable.rs
```diff
@@ -57,75 +46,224 @@ where
...
+    /// Per tick, how much capacity is added to the `Valve`.
+    refill_per_tick: u64,
+}
+
+impl Valve {
+    /// Create a new `Valve` instance
+    fn new(last_tick: u64, maximum_capacity: u64) -> Self {
+        // We do not want a situation where refill never happens. If the maximum
+        // capacity is less than INTERVAL_TICKS we set the floor at 1.
+        let refill_per_tick = cmp::max(1, maximum_capacity / INTERVAL_TICKS);
```
GeorgeHahn, line 83: This retains the ~1MB/sec minimum rate. Is that something that could be addressed here as well? If not, that would be good to document and emit a warning for.

GeorgeHahn (approved):

lading/src/throttle/stable.rs
blt: Hmm, I think I could remove this entirely.

blt: Achieved in [42bc57d](https://github.com/DataDog/lading/pull/660/commits/42bc57dec87592643259942a8d6e7ee3d5b2554b). :) 

The refill is now an f64 so we calculate whatever fractional refill happens per tick, removing the floor of 1 refill per tick or 1Mb/s.

lading_throttle/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "lading-throttle"
```
GeorgeHahn, line 2: This crate will need to be created manually on crates.io (`cargo publish` locally, and then add the team as owners `cargo owner --add github:DataDog:single-machine-performance`)

lading_throttle/src/stable.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// Requested capacity is greater than maximum allowed capacity.
+    #[error("Capacity")]
+    Capacity,
+}
+
+#[derive(Debug)]
+/// A throttle type.
+///
+/// This throttle is stable in that it will steadily refill units at a known
+/// rate and does not inspect the target in any way, compare to `Predictive` in
```
scottopell, line 27: I believe `Predictive` has been removed, this reference doesn't make sense anymore

blt: Good eye, yep.

PR #680, opened 2023-08-17
Remove AsciiStrings use

blt:
### What does this PR do?

This commit removes the use of AsciiStrings struct to generate small strings. Where possible we now use a &str from our `Pool` mechanism. I have also added a number of new benchmarks for most payloads. This unblocks streaming.

### Related issues

REF SMP-664


scottopell (approved):

goxberry (approved):
Very cool.

PR #681, opened 2023-08-17
Use associated types in Generator trait

blt:
### What does this PR do?

Follow-up to #680. I realized that the Generator trait could be used with an associated type to cover owned and reference returns just after #680 merged up. This commit slightly expands the definition of `Generator` and ensures its general use.

### Related issues

REF SMP-664


goxberry (approved):
Ah, yes, that's tidy.

PR #684, opened 2023-08-18
Place block construction in a `Cache` struct

blt:
### What does this PR do?

This commit puts the block construction into a type, called `Cache`. We support a peek interface which requires mutable access and an indexing implementation which is read-only. This is not quite pleasing to the eye but it maintains the interfaces that already exist. This work is intended to be the basis of an infinite streaming `Cache`, not enabled by default. That work will be done in a follow-up PR.

### Additional Notes

For the most part this change is mechanical. The unix-datagram generator now follows the file-gen in sharing a single `Cache` instance per writer, rather than having one for each child.

### Related issues

REF SMP-664

### DIFF
--- a/lading/src/block.rs
+++ b/lading/src/block.rs
@@ -5,16 +5,10 @@ use lading_payload as payload;
 use metrics::gauge;
 use rand::{prelude::SliceRandom, Rng};
 
-#[derive(Debug, PartialEq, Eq, Clone, Copy, thiserror::Error)]
+#[derive(Debug, thiserror::Error, Clone, Copy)]
 pub enum Error {
     #[error("Chunk error: {0}")]
-    Chunk(ChunkError),
-}
-
-impl From<ChunkError> for Error {
-    fn from(error: ChunkError) -> Self {
-        Error::Chunk(error)
-    }
+    Chunk(#[from] ChunkError),
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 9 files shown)


lading/src/block.rs
```diff
@@ -98,113 +79,174 @@ where
...
+        mut rng: &mut R,
+        total_bytes: NonZeroUsize,
+        block_byte_sizes: &[NonZeroUsize],
+        payload: &payload::Config,
+        labels: &Vec<(String, String)>,
+    ) -> Result<Self, Error>
+    where
+        R: Rng + ?Sized,
+    {
+        let block_chunks = chunk_bytes(&mut rng, total_bytes, block_byte_sizes)?;
```
goxberry, line 119: I am concerned that this function call will terminate and return `Ok(_)` in finite time only when `total_bytes` can be expressed as an integral conic combination of the elements of `block_byte_sizes`. That said, as far as I'm aware, none of our generator configuration files violate this criterion.

goxberry: Ah, nevermind; I see the line I missed.

blt: In practice folks put in okay values but there are absolutely values that could be put in to `chunk_bytes` that get goofy results. I'd be very pleased with something that weren't quite so precarious.

blt: Also worth calling out that the algorithm under discussion is unchanged in this PR, so any work to it would be appropriate for a follow-up PR.

goxberry (approved):
Neat.

PR #687, opened 2023-08-22
Block cache with an infinite stream option

blt:
### What does this PR do?

This commit makes two major changes to the block cache. The most important is that the block cache's major CPU consumption -- the construction of `Block` instances -- is now done in a separate OS thread from the tokio runtime. This allows us to introduce the second more important change: infinite streams of `Blocks`. It is now possible for users to construct an unending stream of `Block` instances that do not loop. We maintain a cache of constructed `Block`s up to the maximum total bytes allow to minimize any potential latency impact.

### Additional Notes

Configuration is changed but not in a backward incompatible way.

### Related issues

REF SMP-664

### DIFF
--- a/lading/src/block.rs
+++ b/lading/src/block.rs
@@ -1,20 +1,24 @@
-use std::num::{NonZeroU32, NonZeroUsize};
+use std::{
+    collections::VecDeque,
+    num::{NonZeroU32, NonZeroUsize},
+};
 
 use bytes::{buf::Writer, BufMut, Bytes, BytesMut};
 use lading_payload as payload;
-use metrics::gauge;
-use rand::{prelude::SliceRandom, Rng};
+use rand::{prelude::SliceRandom, rngs::StdRng, Rng, SeedableRng};
+use serde::Deserialize;
+use tokio::sync::mpsc::{self, error::SendError, Sender};
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum SpinError {
+    #[error(transparent)]
+    Send(#[from] SendError<Block>),
+}
 
 #[derive(Debug, thiserror::Error, Clone, Copy)]
 pub enum Error {
     #[error("Chunk error: {0}")]
     Chunk(#[from] ChunkError),
-}
-
-#[derive(Debug)]
-pub(crate) struct Block {
-    pub(crate) total_bytes: NonZeroU32,
-    pub(crate) bytes: Bytes,
 }
 
 #[derive(Debug, thiserror::Error, Clone, Copy)]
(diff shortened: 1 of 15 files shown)


goxberry:
> This allows us to introduce the second more important change: infinite streams of Blocks. It is now possible for users to construct an unending stream of Block instances that do not loop.

I suspect the period of the PRNG used would affect how long it would take before Blocks begin to repeat, though for this application, it seems unlikely we'd exhaust that period unless it were quite small (e.g., for a period of 2^64 * 4 bytes, generating 10^9 bytes/second, around 334 years; I'm guessing that the 1,024 channels aren't generating 10^9 bytes/second each).

blt:
> > This allows us to introduce the second more important change: infinite streams of Blocks. It is now possible for users to construct an unending stream of Block instances that do not loop.
> 
> I suspect the period of the PRNG used would affect how long it would take before Blocks begin to repeat, though for this application, it seems unlikely we'd exhaust that period unless it were quite small (e.g., for a period of 2^64 * 4 bytes, generating 10^9 bytes/second, around 334 years; I'm guessing that the 1,024 channels aren't generating 10^9 bytes/second each).

Ah yeah, maybe better language would be something like "do not repeat from a finite pool of `Block` instances of known size". You are right we're playing the odds here and some values will repeat, but not quite so much as if you generate 1Gb of `Block` instances and loop over that.

lading/src/block.rs
```diff
@@ -170,88 +176,234 @@ impl Cache {
...
+    snd: Sender<Block>,
+) -> Result<(), SpinError> {
+    let mut rng = StdRng::from_seed(seed);
+
+    match payload {
+        payload::Config::TraceAgent(enc) => {
+            let ta = match enc {
+                payload::Encoding::Json => payload::TraceAgent::json(&mut rng),
+                payload::Encoding::MsgPack => payload::TraceAgent::msg_pack(&mut rng),
+            };
```
GeorgeHahn, line 264: Could this be refactored so the configs are handled inside the relevant `payload` implementation? (Is there a future benefit to having access to the payload configs here?)

lading/src/generator/grpc.rs
```diff
@@ -65,6 +74,9 @@ pub struct Config {
     pub block_sizes: Option<Vec<byte_unit::Byte>>,
     /// The maximum size in bytes of the cache of prebuilt messages
     pub maximum_prebuild_cache_size_bytes: byte_unit::Byte,
+    /// Whether to use a fixed or streaming block cache
+    #[serde(default = "crate::block::default_cache_method")]
+    pub block_cache_method: block::CacheMethod,
```
GeorgeHahn, line 79: This kind of seems like it should be set automatically by the payload, would that be possible?

blt: I think it could be but I hadn't considered that. Right now it's the `Cache` that maintains the difference and the payload is really just a mechanism -- to the `Cache` -- of how to get a `Block`. My sense was that we would drive to where streaming was on by default and leave fixed as an option for setups that are especially CPU conscious, so users would rarely see this setting. 

I'm not sure we have enough information to say whether it's appropriate to stream or be fixed automatically, based on payload settings. Maybe I'm missing something you're seeing?

lading/src/block.rs
blt: I think it could be and I started in on that but I felt like it was too big of a change for this PR. I find it awkward, on the subject, that the generator creates the `Cache` at all but I do find it desirable to maintain backward compatibility with our configs.

lading/src/generator/grpc.rs
GeorgeHahn: Hmm, I see. I buy that. We'll have some mismatch if we introduce any streaming-only payloads. We can handle that when we need to though.

GeorgeHahn (approved):

PR #692, opened 2023-08-25
Prefer the use of FxHash over Hash

blt:
### What does this PR do?

This commit adjusts more of our codebase to use the FxHash instead of the randomly seeded Hash{Map|Set}. We want to avoid sources of non-determinism. We believe many of our dependencies will use randomly seeded HashMaps and this commit does nothing to address that.

### Related issues

REF SMP-687

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1057,6 +1057,7 @@ version = "0.1.1"
 dependencies = [
  "prost",
  "prost-build",
+ "rustc-hash",
  "serde",
  "serde_json",
  "uuid",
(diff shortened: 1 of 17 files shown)


GeorgeHahn (approved):

goxberry:
For this sort of change, I'm wondering if we should be doing another kind of analysis. If we make this change to decrease non-determinism -- which I agree is preferable -- is that decrease something we should be quantifying in some meaningful way? (Decreased variance in throughput? Change in distribution of throughputs?)

blt:
> For this sort of change, I'm wondering if we should be doing another kind of analysis. If we make this change to decrease non-determinism -- which I agree is preferable -- is that decrease something we should be quantifying in some meaningful way? (Decreased variance in throughput? Change in distribution of throughputs?)

I think this is a really important question and it's not something I have an answer for. But I would like to.

PR #698, opened 2023-09-11
Allow the user to configure the range of DogStatsD metrics

blt:
### What does this PR do?

This commit introduces a configuration option to allow the user to define the range of values that appear in DogStatsD metrics. This range applies to all metric kinds. The distribution is changed from Standard -- "numerically uniform" -- to actually Uniform. We do not allow users to configure the distribution.

### Related issues

REF SMP-694

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.18.1-rc2]
+### Fixed
+- DogStatsD value min/max is now 2**63 symmetric. This avoids a range issue in the `rand` crate.
+
 ## [0.18.1-rc1]
 ### Added
 - It is now possible for users to configure the range of DogStatsD payloads
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,35 +6,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-## [0.18.1-rc6]
-### Changed
-- The configuration option `float_weight` is changed to `float_probability`.
-
-## [0.18.1-rc5]
-### Changed
-- Capitalization is renamed to snake-case in value configuration.
-
-## [0.18.1-rc4]
-### Changed
-- The int-only experiment is now reverted.
-- Configuration for numeric values now explicitly allows constant values.
-
-## [0.18.1-rc3]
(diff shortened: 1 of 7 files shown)


GeorgeHahn (approved):

lading_payload/src/dogstatsd.rs
```diff
@@ -27,6 +27,14 @@ fn contexts_maximum() -> u16 {
     10_000
 }
 
+fn value_minimum() -> i64 {
+    i64::MIN
```
scottopell, line 31: I'd vote for more reasonable default values, this is known to be a worst-case for sketch calculation, right?

CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.18.1-rc3]
+### Changed
+- DogStatsD NumValue is now int-only for experimental purposes, range is also made inclusive. The inclusivity will remain.
```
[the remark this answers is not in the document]
blt: Correct.

lading_payload/src/dogstatsd.rs
blt: We now believe that the range is only part of it. If you send `1` rapidly enough you can force the distribution to allocate like mad. @GeorgeHahn will have a writeup when he's back from PTO.

lading_payload/src/dogstatsd.rs
```diff
@@ -118,6 +128,30 @@ impl Default for MetricWeights {
     }
 }
 
+/// Configuration for the values of a metric.
+#[derive(Debug, Deserialize, Clone, PartialEq, Copy)]
+pub struct ValueConf {
+    /// Odds out of 256 that the value will be a float and not an integer.
+    float_weight: u8,
```
scottopell, line 135: This is a fairly unintuitive unit for "chance of thing occurring" (I came here from https://github.com/DataDog/single-machine-performance/pull/1268/files#diff-34dd14dde2466a88f431ddc27fc1d897ffede2f88d81eb37711a0d5886b0af92R29), especially given the prior art of `multipack_value_probability`.
What do you think of making this a 0-1 value and renaming to `probability`?

(Also open to renaming/changing `multipack_value_probability`, just looking for consistency between these two options really)

blt: Ah, that's a good call-out. I'd rather this be consistent.

PR #706, opened 2023-10-16
Introduce 'idle' generator

blt:
### What does this PR do?

This commit introduces a generator that is idle. We don't allow for empty generator lists but do want to be able to express experiments where lading produces no load into a target. An alternative would be to allow for 0 bytes/sec in all generators which I find undesirable.

### Related issues

REF SMPTNG-23


blt:
Closing in favor of empty generator lists in configuration.

PR #717, opened 2023-10-20
Corrects example dogstatsd configuration

scottopell:
### What does this PR do?

Previous example dogstatsd config had a bug where weights of metric_kinds added up to greater than a `u8`.
Corrects some of the specific values as well.


GeorgeHahn:
Is it invalid for the sum of the weights to be larger than `u8::MAX`? I spot checked the code and didn't see an issue there.

goxberry (approved):
Makes sense; probabilities should be less than or equal to 1.

PR #720, opened 2023-11-02
Introduce 'logrotate' log generator

blt:
### What does this PR do?

This commit introduces a new log generator that is able to mimic logrotate's operation. This also allows us to mimic how kubernetes logs work by setting `total_rotations` to be zero.

### Related issues

REF SMPTNG-44

### DIFF
--- a/lading/src/generator/file_gen/logrotate.rs
+++ b/lading/src/generator/file_gen/logrotate.rs
@@ -16,7 +16,7 @@
 //!
 use std::{
     num::{NonZeroU32, NonZeroUsize},
-    path::PathBuf,
+    path::{Path, PathBuf},
     str, thread,
 };
 
@@ -175,7 +175,7 @@ impl Server {
             basename.set_extension("log");
 
             let child = Child::new(
-                basename,
+                &basename,
                 config.total_rotations,
                 bytes_per_second,
                 maximum_bytes_per_file,
@@ -229,7 +229,7 @@ struct Child {
 
 impl Child {
     fn new(
-        basename: PathBuf,
+        basename: &Path,
         total_rotations: u8,
         bytes_per_second: NonZeroU32,
         maximum_bytes_per_log: NonZeroU32,
@@ -241,13 +241,10 @@ impl Child {
         for i in 0..total_rotations {
             let name = format!(
                 "{orig}.{i}",
-                orig = basename
-                    .file_name()
-                    .unwrap_or_default()
-                    .to_string_lossy()
-                    .to_string()
+                orig = basename.file_name().unwrap_or_default().to_string_lossy()
             );
-            let mut pth = basename.clone();
+            let mut pth = PathBuf::new();
+            pth.push(basename);
             pth.set_file_name(name);
             names.push(pth);
         }
--- a/lading/src/generator/file_gen/logrotate.rs
+++ b/lading/src/generator/file_gen/logrotate.rs
@@ -1,0 +1,350 @@
+//! The lading 'logrotate' file generator.
+//!
+//! The logrotate file generator does not "connect" however losely to the target
+//! but instead, without coordination, merely writes files on disk. We mimic the
+//! fashion in which logrotate operates. All log files are written up to some
+//! maximum amount, with a fixed number of 'rotations' happening per file. Once
+//! a file reaches its maximum size its name is removed although writes may
+//! still arrive after the name removal.
+//!
+//! ## Metrics
+//!
+//! `bytes_written`: Total bytes written
+//! `bytes_per_second`: Configured rate to send data
+//!
+//! Additional metrics may be emitted by this generator's [throttle].
+//!
+use std::{
+    num::{NonZeroU32, NonZeroUsize},
+    path::PathBuf,
+    str, thread,
+};
+
+use byte_unit::{Byte, ByteUnit};
+use futures::future::join_all;
+use lading_throttle::Throttle;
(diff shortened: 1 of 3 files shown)


lading/src/generator/file_gen/logrotate.rs
```diff
@@ -0,0 +1,350 @@
+//! The lading 'logrotate' file generator.
+//!
+//! The logrotate file generator does not "connect" however losely to the target
```
goxberry, line 3: losely -> loosely

blt: Addressed in [9f3d917](https://github.com/DataDog/lading/pull/720/commits/9f3d9171149b4741f4de5ffb2beb9a1438bd6caf)

lading/src/generator/file_gen/logrotate.rs
```diff
@@ -0,0 +1,347 @@
...
+
+    async fn spin(mut self) -> Result<(), Error> {
+        let bytes_per_second = self.bytes_per_second.get() as usize;
+        let mut total_bytes_written: u64 = 0;
+        let maximum_bytes_per_log: u64 = u64::from(self.maximum_bytes_per_log.get());
+
+        let total_names = self.names.len();
+        let last_name = &self.names[total_names];
+
+        // SAFETY: By construction the name is guranteed to have a parent.
```
goxberry, line 270: guranteed -> guaranteed

lading/src/generator/file_gen/logrotate.rs
```diff
@@ -0,0 +1,347 @@
...
+        // Move the block_cache into an OS thread, exposing a channel between it
+        // and this async context.
+        let block_cache = self.block_cache;
+        let (snd, rcv) = mpsc::channel(1024);
+        let mut rcv: PeekableReceiver<Block> = PeekableReceiver::new(rcv);
+        thread::Builder::new().spawn(|| block_cache.spin(snd))?;
+        let bytes_written = register_counter!("bytes_written");
+
+        loop {
+            let blk = rcv.peek().await.unwrap();
```
goxberry, line 291: Is it worth having a safe unwrap comment here, or is it meant to be understood that the block cache never emits `None` from its `spin` method by virtue of its construction?

lading/src/generator/file_gen/logrotate.rs
```diff
@@ -0,0 +1,347 @@
...
+        }
+    }
+
+    async fn spin(mut self) -> Result<(), Error> {
+        let bytes_per_second = self.bytes_per_second.get() as usize;
+        let mut total_bytes_written: u64 = 0;
+        let maximum_bytes_per_log: u64 = u64::from(self.maximum_bytes_per_log.get());
+
+        let total_names = self.names.len();
+        let last_name = &self.names[total_names];
```
goxberry, line 268: Should the right-hand side of this statement be `&self.names[total_names - 1]`?

goxberry:
Is this PR still work in progress? It looks as though it isn't yet possible to configure a logrotate file generator via a configuration file.

goxberry:
Otherwise, this is looking good to me. Very cool.

blt:
> Otherwise, this is looking good to me. Very cool.

It is. Seems like the basic idea lands, today I intend to wire it up and try it out on some RCs.

lading/src/generator/file_gen/logrotate.rs
blt: `len` is zero-index, so I think this works as-is. I also regularly forget that https://doc.rust-lang.org/std/primitive.slice.html#method.last exists, so I'll use that instead.

lading/src/generator/file_gen/logrotate.rs
blt: It could be more clear; I'll add a safety comment.

lading/src/generator/file_gen/logrotate.rs
goxberry: Isn't the zero-indexing convention why this code will panic? For instance, the code below ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3a22fe7476670701ddf3d0ac9b861787)) will panic on a bounds check:

```rust
fn main() {
    let v = vec![0, 1, 2, 3];
    let len = v.len();
    println!("v[{len}] = {x}", x = &v[len]);
}
```

blt: Oh yeah, of course. I wasn't paying close attention but you're absolutely right.

blt: Addressed in [5a44458](https://github.com/DataDog/lading/pull/720/commits/5a444584eddc8889a28778a0acb13e34d78d8762)

lading/src/generator/file_gen/logrotate.rs
blt: Addressed in [5a44458](https://github.com/DataDog/lading/pull/720/commits/5a444584eddc8889a28778a0acb13e34d78d8762)

lading/src/generator/file_gen/logrotate.rs
blt: Addressed in [5a44458](https://github.com/DataDog/lading/pull/720/commits/5a444584eddc8889a28778a0acb13e34d78d8762)

goxberry (approved):

GeorgeHahn (approved):

PR #721, opened 2023-11-08
Adjust proofs for stable valve in throttle

blt:
### What does this PR do?

This commit adjusts the proofs that we have in our stable throttle to avoid a loop and instead rely on a series of smaller proofs. This terminates in a reasonable amount of time. Note I've removed the proptests and this commit will need to be followed by another than runs kani in CI.

I may still be missing a few small properties.

### Related issues

REF SMPTNG-79

### DIFF
--- a/lading_throttle/src/stable.rs
+++ b/lading_throttle/src/stable.rs
@@ -98,16 +98,16 @@ impl Valve {
     #[allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
     fn request(&mut self, ticks_elapsed: u64, capacity_request: u32) -> Result<u64, Error> {
         // Okay, here's the idea. We have bucket that fills every INTERVAL_TICKS
-        // seconds up and requests draw down on that bucket. When it's empty, we
-        // return the number of ticks until the next interval roll-over. Callers
-        // are expected to wait although nothing forces them to. Capacity is
-        // only drawn on when it is immediately available.
-        if capacity_request == 0 {
-            return Ok(0);
-        }
-
-        // Fast bail-out. There's no way for this to ever be satisfied and is a
-        // bug on the part of the caller, arguably.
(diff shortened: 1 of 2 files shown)


GeorgeHahn (approved):

lading_throttle/src/stable.rs
```diff
@@ -121,167 +121,121 @@ impl Valve {
...
+            valve.capacity == maximum_capacity - request,
+            "Request in-capacity, interval should reduce capacity by request size.",
+        );
+    }
 
-        let iters: usize = kani::any();
-        kani::assume(iters < 100);
+    /// If a request is made on the throttle such that capacity > request <
+    /// max_capacity and ticks_elapsed <= INTERVAL_TICKS then the request should
+    /// return with non-zero slop and the internal capacity of the valve should
```
goxberry, line 201: It seems like the capacity inequality should read `capacity < request <= max_capacity`, based on the proof below.

lading_throttle/src/stable.rs
```diff
@@ -121,167 +121,121 @@ impl Valve {
...
+        let mut valve = Valve::new(maximum_capacity);
+        let maximum_capacity = maximum_capacity.get();
 
-        let mut slop = 0;
-        for _ in 0..iters {
-            let request: NonZeroU32 = kani::any();
-            kani::assume(request.get() <= maximum_capacity);
+        let original_capacity = valve.capacity;
+        let request: u32 =
+            kani::any_where(|r: &u32| *r <= maximum_capacity && *r > original_capacity);
```
goxberry, line 211: Minor nit: would favor writing as `original_capacity < *r && *r <= maximum_capacity` to make it more obvious that `*r` must be between those values (upper bound inclusive).

goxberry (approved):

PR #723, opened 2023-11-08
Convert target metrics where possible to stream

blt:
### What does this PR do?

This commit converts our expvar and prometheus metrics to the use of streaming bodies where possible, especially on the prometheus side. The implementation of both is adjusted to be similar and we now record the content-length from the target.

### Related issues

REF SMPTNG-79


GeorgeHahn (approved):

goxberry (approved):

blt:
While I think it is an interesting bit of work with some debugging around failure modes it's also unrelated to SMPTNG-79.

PR #725, opened 2023-11-09
Introduce proofs for block mechanism

blt:
### What does this PR do?

This commit adds proofs for the `chunk_bytes` function, converting its interface to be in terms of `u32` -- as matches our configuration -- and also to use a round-robin method internally to avoid randomness. The callers -- two sites -- are responsible for passing  `chunks` in as a mutable reference, allowing us to bound the maximum iterations of this function. The function is now allocation-free, although the change as a whole is neutral in this regard. 

### Motivation

REF #724
REF SMPTNG-79

### DIFF
--- a/lading_payload/src/block.rs
+++ b/lading_payload/src/block.rs
@@ -5,10 +5,7 @@
 //! from that, decoupling the create/send operations. This module is the
 //! mechanism by which 'blocks' -- that is, byte blobs of a predetermined size
 //! -- are created.
-use std::{
-    collections::VecDeque,
-    num::{NonZeroU32, NonZeroUsize},
-};
+use std::{collections::VecDeque, num::NonZeroU32};
 
 use bytes::{buf::Writer, BufMut, Bytes, BytesMut};
 use rand::{prelude::SliceRandom, rngs::StdRng, Rng, SeedableRng};
@@ -91,9 +88,9 @@ pub enum Cache {
         /// The seed used to construct the `Cache`.
         seed: [u8; 32],
         /// The total number of bytes that will be generated.
-        total_bytes: usize,
+        total_bytes: u32,
         /// The sizes of the blocks that will be generated.
-        block_chunks: Vec<usize>,
+        block_chunks: Vec<u32>,
         /// The payload that will be generated.
         payload: crate::Config,
     },
@@ -112,13 +109,11 @@ impl Cache {
     /// of `block_byte_sizes` is large than `total_bytes`.
     pub fn stream(
         seed: [u8; 32],
-        total_bytes: NonZeroUsize,
-        block_byte_sizes: &[NonZeroUsize],
+        total_bytes: NonZeroU32,
+        block_byte_sizes: &[NonZeroU32],
         payload: crate::Config,
     ) -> Result<Self, Error> {
-        let mut rng = StdRng::from_seed(seed);
-
-        let block_chunks = chunk_bytes(&mut rng, total_bytes, block_byte_sizes)?;
+        let block_chunks = chunk_bytes(total_bytes, block_byte_sizes)?;
         Ok(Self::Stream {
             seed,
             total_bytes: total_bytes.get(),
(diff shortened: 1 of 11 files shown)


GeorgeHahn (approved):

lading_payload/src/block.rs
```diff
@@ -520,62 +537,115 @@ where
...
+        let total_bytes: NonZeroU32 = kani::any();
+        let mut block_chunks: [u32; 10] = [0; 10];
+
+        let under: NonZeroU32 = kani::any_where(|x| *x < total_bytes);
+        let equal = total_bytes;
+        let over: NonZeroU32 = kani::any_where(|x| *x > total_bytes);
+
+        kani::assert(
+            chunk_bytes(total_bytes, &[under, equal, over], &mut block_chunks).is_err(),
+            "chunk_bytes should fail when all sizes not all under the limit",
```
goxberry, line 604: "chunk_bytes should fail when not all sizes under the limit"?

lading_payload/src/block.rs
```diff
@@ -392,21 +381,49 @@ where
...
+            let block_size = block_size.get();
+            if block_size <= bytes_remaining {
+                chunks[total_chunks] = block_size;
+                total_chunks += 1;
+                bytes_remaining -= block_size;
+            }
+        }
+    }
+
+    Ok(total_chunks)
```
goxberry, line 426: Let `k` be the largest index such that:

- `0 <= k < N`
- `chunks[..k].iter().sum() <= total_bytes`

Should every element of `chunks[k..]` be zero?

blt: Oooh yes they should.

blt: Proof added in [eb556e1](https://github.com/DataDog/lading/pull/725/commits/eb556e1b9bd49d1b212a97d1d66827e8da6a1e29).

lading_payload/src/block.rs
blt: Modified in [eb556e1](https://github.com/DataDog/lading/pull/725/commits/eb556e1b9bd49d1b212a97d1d66827e8da6a1e29).

lading_payload/src/block.rs
```diff
@@ -359,30 +359,19 @@ fn stream_inner(
...
-    block_byte_sizes: &[NonZeroUsize],
-) -> Result<Vec<usize>, Error>
-where
-    R: Rng + Sized,
-{
+fn chunk_bytes<const N: usize>(
+    total_bytes: NonZeroU32,
+    block_byte_sizes: &[NonZeroU32],
+    chunks: &mut [u32; N],
+) -> Result<usize, Error> {
```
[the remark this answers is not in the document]
blt: In practice these are all the size `MAX_CHUNKS` so we could possibly just substitute N for that. The proofs would take longer to run -- note the proofs are in terms of N == 10 -- and also if we want to allow streams and fixed caches to have different maxes we can achieve that now. 

A little bit it makes the proof run better, a little bit it makes the code easier to modify in the future but we don't strictly speaking need to.

lading_payload/src/block.rs
```diff
@@ -359,30 +359,19 @@ fn stream_inner(
...
-/// up correctly is not necessarily straightforward. This utility method does
-/// the computation in cases where it would otherwise be annoying. From the
-/// allowable block sizes -- defined by `block_byte_sizes` -- a random member is
-/// chosen and is deducted from the total bytes remaining. This process
-/// continues until the total bytes remaining falls below the smallest block
-/// size. It's possible that a user could supply just the right parameters to
-/// make this loop infinitely. A more clever algorithm would be great.
+/// We partition `total_bytes` by `block_byte_sizes` via a round-robin method.
+/// Our goal is to terminate, more or less partition the space and do so without
+/// biasing toward any given byte size.
```
goxberry, line 364: Is determinism of the `chunks` configuration upon output a desirable property when used as a load generator? I see the appeal from the perspective of writing proofs that will execute quickly -- and narrowing down debugging surface -- I'm just wondering if the deterministic round-robin strategy is something we would want to use for a situation other than testing.

blt: It is. Worth recalling that even though the previous implementation took a rand the output was always deterministic, since we seed lading's RNG deterministically. The round-robin is just a new idea I had to get a deterministic output but that specific part I don't want to advertise out to users. Determinism yes, but how we achieve that is up to us.

goxberry (approved):

PR #727, opened 2023-11-10
Introduce fuzzing for `lading_payload::block::Cache::fixed`

blt:
### What does this PR do?

This commit introduces a new fuzzing harness for the above mentioned function. We want to ensure that we are not crashing at runtime or consuming more memory than allowed.

### Related issues

REF SMPTNG-79

PR #735, opened 2023-11-15
Convert the generation of tagsets to at-time

blt:
### What does this PR do?

This commit modifies the tag generator to generate tagsets at-time. The way we maintain context constraints is novel in the project: we maintain an internal RNG that we deterministically reseed when we loop around the number of contexts.

### Related issues

Follow-up from #727.
REF SMPTNG-79

### DIFF
--- a/lading_payload/src/dogstatsd/common/tags.rs
+++ b/lading_payload/src/dogstatsd/common/tags.rs
@@ -1,48 +1,91 @@
-use std::rc::Rc;
+use std::{
+    cell::{Cell, RefCell},
+    rc::Rc,
+};
+
+use rand::{rngs::SmallRng, SeedableRng};
 
 use crate::{common::strings, dogstatsd::ConfRange};
 
 // This represents a list of tags that will be present on a single
 // dogstatsd message.
 pub(crate) type Tagset = Vec<String>;
-// Multiple tagsets. Useful to generate as a batch (is it??)
-pub(crate) type Tagsets = Vec<Tagset>;
 
+/// Generator for tags
+///
+/// This is an unusual generator. Unlike our others this maintains its own RNG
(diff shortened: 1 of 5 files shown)


lading_payload/src/dogstatsd/common/tags.rs
```diff
@@ -1,48 +1,91 @@
...
 pub(crate) type Tagset = Vec<String>;
-// Multiple tagsets. Useful to generate as a batch (is it??)
-pub(crate) type Tagsets = Vec<Tagset>;
 
+/// Generator for tags
+///
+/// This is an unusual generator. Unlike our others this maintains its own RNG
+/// and will reseed it when counter reaches `num_tagsets`. The goal is to
+/// produce only a limited, deterministic set of tags while avoiding needing to
+/// allocate them all in one shot.
```
goxberry, line 19: If the goal of this implementation is to produce a limited, deterministic set of tags, how much value do we get from reseeding this generator's RNG once it reaches `num_tagsets`?

goxberry:
What does "at-time" mean in the description of this PR?

goxberry (approved):
Cool! I'm interested in seeing the memory behavior of this PR in production.

lading_payload/src/dogstatsd/common/tags.rs
blt: The generator is responsible for producing an infinite amount of tagsets, it's just that the tagsets have to start repeating after we've seen `num_tagsets` of them. Previously this was achieved by having the generator produce that finite set in one shot and then the caller was responsible for sampling in the finite set. 'Finite' is a load bearing word here and can be practically too large.

blt:
> What does "at-time" mean in the description of this PR?

Ah, at the time of use.

PR #739, opened 2023-11-20
Sample rate config for dogstatsd

GeorgeHahn:
### What does this PR do?

Add sample rate & sample rate probability configs to the Dogstatsd payload generator.

### Motivation

This parameter was previously randomly generated between `0..=1`, and we'd like more control over it.

### Related issues

SMPTNG-72


lading_payload/src/dogstatsd.rs
```diff
@@ -258,6 +266,15 @@ pub struct Config {
     #[serde(default = "multivalue_count")]
     pub multivalue_count: ConfRange<u16>,
 
+    /// Sampling rate to report for metrics
```
scottopell, line 269:
```suggestion
    /// Range of possible values for the sampling rate sent in dogstatsd messages
```

lading_payload/src/dogstatsd.rs
```diff
@@ -258,6 +266,15 @@ pub struct Config {
...
+    #[serde(default = "sampling")]
+    pub sampling: ConfRange<f32>,
```
scottopell, line 271:
```suggestion
    pub sampling_range: ConfRange<f32>,
```

lading_payload/src/dogstatsd.rs
```diff
@@ -258,6 +266,15 @@ pub struct Config {
...
+
+    /// Probability between 0 and 1 that a given dogstatsd msg
```
scottopell, line 273:
```suggestion
    /// Probability between 0 and 1 that a given dogstatsd msg will specify a sampling rate.
    /// The sampling rate is chosen from `sampling_range`
```

lading_payload/src/dogstatsd.rs
```diff
@@ -258,6 +266,15 @@ pub struct Config {
...
+    /// will be reported as a sampled value.
```
scottopell, line 274:
```suggestion
```

lading_payload/src/dogstatsd/common.rs
```diff
@@ -100,6 +100,27 @@ pub(crate) enum ZeroToOne {
...
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum ZeroToOneError {
+    OutOfRange,
+}
+
+impl TryFrom<f32> for ZeroToOne {
+    type Error = ZeroToOneError;
+
+    fn try_from(value: f32) -> Result<Self, Self::Error> {
+        if (value - 1.0).abs() < f32::EPSILON {
```
goxberry, line 112: The value 1.0 is exactly representable in IEEE-754-conformant 32-bit floating-point arithmetic, so the epsilon check here should be unnecessary, though I appreciate the intent because it is usually good practice.

lading_payload/src/dogstatsd/metric.rs
```diff
@@ -85,11 +91,15 @@ impl<'a> Generator<'a> for MetricGenerator {
         let template: &Template = self.templates.choose(&mut rng).unwrap();
 
         let container_id = choose_or_not_ref(&mut rng, &self.container_ids).map(String::as_str);
-        // TODO sample_rate should be option and have a probability that determines if its present
-        // Mostly inconsequential for the Agent, for certain metric types the Agent
-        // applies some correction based on this value. Affects count and histogram computation.
         // https://docs.datadoghq.com/metrics/custom_metrics/dogstatsd_metrics_submission/#sample-rates
-        let sample_rate = rng.gen();
+        let prob: f32 = OpenClosed01.sample(&mut rng);
+        let sample_rate = if prob < self.sampling_probability {
```
goxberry, line 96: Minor pedantic nit: because floating-point arithmetic quantizes the range [0, 1], omitting zero here does have a small effect on the probability logic that would be material for small values of `self.sampling_probability` (e.g., `f64::MIN_POSITIVE`). Does this actually matter in practice? Probably not, unless our users are pedantic masochists.

blt (approved):

lading_payload/src/dogstatsd/metric.rs
GeorgeHahn: Merged, but I'd still like to understand this better. Would one of the other provided distributions be more correct? https://docs.rs/rand/latest/rand/distributions/struct.Standard.html#floating-point-implementation The `Standard` distribution stands out to me, but it seems like it would have the same issue on the opposite side.

PR #748, opened 2023-11-27
Use generational storage in the CaptureManager

blt:
### What does this PR do?

This commit makes use of generational storage in the CaptureManager, expiring keys that have not been written to for ten seconds. After this period the keys will no longer be written out. This is especially important for observer metrics in the presence of short-lived processes.

### Related issues

REF SMPTNG-85

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1073,6 +1073,7 @@ dependencies = [
  "procfs",
  "proptest",
  "proptest-derive",
+ "quanta",
  "rand",
  "reqwest",
  "rustc-hash",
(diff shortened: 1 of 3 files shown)


GeorgeHahn (approved):

tobz (approved):
This seems correct to me from a quick pass. 👍🏻 

One bit to note that is that, currently, expiring metrics is incompatible with using metric handles. This is to say that, if you're only ever doing `counter!(....`), you're fine... but if you do `let counter = counter!(...)` and then store/work with that counter (or gauge, etc) handle later on, and _then_ the metric that handle represents is marked as expired... the handle effectively dies for the remainder of the process. You would need to re-register it specifically... which is why the normal `counter!(...)` usage is fine, because it always grabs a new handle (which re-registers it if need be).

PR #752, opened 2023-11-29
Introduce 'default tags' to the dogstatsd payload

blt:
### What does this PR do?

This commit allows the user to specify a list of default tags to prefix on any generated tagset. These tags are assumed to be in valid form -- key:value,key:value -- without a trailing comma. Users may pass an empty string.

### Related issues

REF SMPTNG-91

### DIFF
--- a/lading_payload/src/dogstatsd.rs
+++ b/lading_payload/src/dogstatsd.rs
@@ -86,6 +86,10 @@ fn sampling_probability() -> f32 {
     0.5
 }
 
+fn default_tags() -> Vec<String> {
+    Vec::new()
+}
+
 /// Weights for `DogStatsD` kinds: metrics, events, service checks
 ///
 /// Defines the relative probability of each kind of `DogStatsD` datagram.
@@ -258,7 +262,7 @@ where
 }
 
 /// Configure the `DogStatsD` payload.
-#[derive(Debug, Deserialize, Clone, PartialEq, Copy)]
+#[derive(Debug, Deserialize, Clone, PartialEq)]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 #[serde(deny_unknown_fields)]
 pub struct Config {
@@ -287,6 +291,11 @@ pub struct Config {
     /// separated by a :
     #[serde(default = "tags_per_msg")]
     pub tags_per_msg: ConfRange<u8>,
+
+    /// Selection of tag pairs to always prepend to tagsets. Assumed to be
+    /// valid, do not count toward `tags_per_msg`.
+    #[serde(default = "default_tags")]
+    pub default_tags: Vec<String>,
 
     /// Probability between 0 and 1 that a given dogstatsd msg
     /// contains multiple values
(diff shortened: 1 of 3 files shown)


lading_payload/src/dogstatsd.rs
```diff
@@ -288,6 +292,11 @@ pub struct Config {
     #[serde(default = "tags_per_msg")]
     pub tags_per_msg: ConfRange<u8>,
 
+    /// Selection of tag pairs to always prepend to tagsets. Assumed to be
+    /// valid, do not count toward `tags_per_msg`.
+    #[serde(default = "default_tags")]
+    pub default_tags: Vec<String>,
```
GeorgeHahn, line 298: Optional: this could be validated using serde's `deserialize_with` attribute https://serde.rs/field-attrs.html#deserialize_with

lading_payload/src/dogstatsd/common/tags.rs
```diff
@@ -61,6 +64,19 @@ impl<'a> crate::Generator<'a> for Generator {
     {
         let mut tagset = Vec::new();
 
+        if let Some(default_tag) = self
+            .default_tags
+            .choose(&mut *self.internal_rng.borrow_mut())
```
GeorgeHahn, line 69: I may be misunderstanding how this works; this looks like it'd pick one tag to prepend. Is that correct? That behavior doesn't match the config doc string.

scottopell:
I don't think prepending a static set of tags will meaningfully change the relationship between # contexts and # unique tags, which is the goal of SMPTNG-91.

I don't have a super clear idea of how I want to do this, but something along these lines:
1. Generate a "small" amount of tags to be the "common tags" (tbd what "small" means, maybe num_contexts / 100 ?)
2. For each desired context, sample out of "common tags" until we reach a desired "common" amount (maybe num_tags / 2 ?)
3. At this point, our metric template may or may not be a unique context (we could have chose the same "common tags" in step 2 for some other metric template). Fill out the rest of the tags with random & unique tags to get a unique context.

There is some uncertainty in there, but I think the two key distinctions I want to make is that the "common" tags are sampled from a pool (rather than being a static set) and that they consume from the "num_desired_tags" budget.

blt:
> Generate a "small" amount of tags to be the "common tags" (tbd what "small" means, maybe num_contexts / 100 ?)

I'd vote for a configurable value. 

> There is some uncertainty in there, but I think the two key distinctions I want to make is that the "common" tags are sampled from a pool (rather than being a static set) and that they consume from the "num_desired_tags" budget.

Gotcha. Okay, that would be a little more involved then. It's not a hard pull -- you have to make a pool as a part of the tag generator that's sampled from preferentially -- but this PR isn't that.

scottopell:
> Gotcha. Okay, that would be a little more involved then. It's not a hard pull -- you have to make a pool as a part of the tag generator that's sampled from preferentially -- but this PR isn't that.

fair enough, if you'd like to land this as-is I don't object, I just don't think it fully addresses the gap described in smptng-91.

scottopell:
its still a WIP, but I've started to implement my ideas in https://github.com/DataDog/lading/pull/761

blt:
Closing in favor of #761.

PR #759, opened 2023-12-04
Waits on generator's execution in order to print out any errors

scottopell:
### What does this PR do?
Adds logic in the main `select` loop of `lading` to wait for the result of the generators that were spawned.

### Motivation
While there is good error handling inside the generators that give feedback when an invalid condition is hit, that result was not being read anywhere.

### DIFF
--- a/lading/src/bin/lading.rs
+++ b/lading/src/bin/lading.rs
@@ -295,6 +295,7 @@ async fn inner_main(
 
     let (tgt_snd, _tgt_rcv) = broadcast::channel(1);
 
+    let mut gsrv_handles: Vec<tokio::task::JoinHandle<Result<(), generator::Error>>> = Vec::new();
     //
     // GENERATOR
     //
@@ -302,7 +303,11 @@ async fn inner_main(
         let tgt_rcv = tgt_snd.subscribe();
         let generator_server = generator::Server::new(cfg, shutdown.clone()).unwrap();
         let _gsrv = tokio::spawn(generator_server.run(tgt_rcv));
+        gsrv_handles.push(_gsrv);
     }
+    // select_all will yield the first resolved future in the given iterator
+    // It will continue to yield new futures as they resolve.
+    let gsrv_waitable = futures::future::select_all(gsrv_handles);
 
     //
     // INSPECTOR
@@ -384,6 +389,13 @@ async fn inner_main(
         _ = experiment_sleep => {
             info!("experiment duration exceeded, signaling for shutdown");
             shutdown.signal().unwrap();
+        }
+        (item_resolved, ready_future_index, _remaining_futures) = gsrv_waitable => {
+            match item_resolved {
+                Ok(Ok(())) => debug!("generator {ready_future_index} shut down successfully"),
+                Ok(Err(err)) => error!("generator failed with err:\n{}", err),
+                Err(join_err) => error!("generator failed with unknown join err:\n{}", join_err),
+            }
         }
         res = tsrv => {
             match res {
--- a/lading/src/bin/lading.rs
+++ b/lading/src/bin/lading.rs
@@ -295,19 +295,15 @@ async fn inner_main(
 
     let (tgt_snd, _tgt_rcv) = broadcast::channel(1);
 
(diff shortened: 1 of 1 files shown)


scottopell:
This is a slightly different pattern from the other "waitables" here, for example the blackholes spawn an async task in-line that blocks on the results, rather than wait in the central `select`.

Is there any significant advantage to one vs the other? The `future::select_all` seems preferable to me with the goal of having fewer tasks, but I'm not too familiar with tokio and if this is a valid consideration.

lading/src/bin/lading.rs
```diff
@@ -295,14 +295,19 @@ async fn inner_main(
...
 
+    let mut gsrv_handles: Vec<tokio::task::JoinHandle<Result<(), generator::Error>>> = Vec::new();
     //
     // GENERATOR
     //
     for cfg in config.generator {
         let tgt_rcv = tgt_snd.subscribe();
         let generator_server = generator::Server::new(cfg, shutdown.clone()).unwrap();
         let _gsrv = tokio::spawn(generator_server.run(tgt_rcv));
+        gsrv_handles.push(_gsrv);
```
blt, line 306: As a matter of convention this should be changed to `gsrv` here and the line prior.

lading/src/bin/lading.rs
```diff
@@ -295,14 +295,19 @@ async fn inner_main(
...
     }
+    // select_all will yield the first resolved future in the given iterator
+    // It will continue to yield new futures as they resolve.
+    let gsrv_waitable = futures::future::select_all(gsrv_handles);
```
blt, line 310: I'd prefer the use of a [`JoinSet`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html). While I think this will work keeping all our uses to tokio avoids historical performance stutters caused by different assumptions between futures/tokio and means upgrades to the crates come for free.

lading/src/bin/lading.rs
```diff
@@ -385,6 +390,13 @@ async fn inner_main(
             info!("experiment duration exceeded, signaling for shutdown");
             shutdown.signal().unwrap();
         }
+        (item_resolved, ready_future_index, _remaining_futures) = gsrv_waitable => {
+            match item_resolved {
+                Ok(Ok(())) => debug!("generator {ready_future_index} shut down successfully"),
+                Ok(Err(err)) => error!("generator failed with err:\n{}", err),
+                Err(join_err) => error!("generator failed with unknown join err:\n{}", join_err),
+            }
+        }
```
blt, line 399: Neat.

blt (changes requested):

blt:
> This is a slightly different pattern from the other "waitables" here, for example the blackholes spawn an async task in-line that blocks on the results, rather than wait in the central `select`.
> 
> Is there any significant advantage to one vs the other? The `future::select_all` seems preferable to me with the goal of having fewer tasks, but I'm not too familiar with tokio and if this is a valid consideration.

The major difference is that if there is a failure in the blackhole the main task will not be aware of it. This is probably not the best construction. We may also, now that I'm thinking about it, want to bubble the generator error up in an error return, shutting the program down.

lading/src/bin/lading.rs
scottopell: Thanks! I didn't run across `JoinSet`, it seems like a better fit given that these handles are all already `JoinHandles`.

scottopell:
> The major difference is that if there is a failure in the blackhole the main task will not be aware of it. This is probably not the best construction. We may also, now that I'm thinking about it, want to bubble the generator error up in an error return, shutting the program down.

Would probably be a good future pass through this logic, to ensure that errors in any subcomponent shut down the program (rather than simply log an error as the current implementation does).

blt:
> > The major difference is that if there is a failure in the blackhole the main task will not be aware of it. This is probably not the best construction. We may also, now that I'm thinking about it, want to bubble the generator error up in an error return, shutting the program down.
> 
> Would probably be a good future pass through this logic, to ensure that errors in any subcomponent shut down the program (rather than simply log an error as the current implementation does).

Strong agree.

PR #761, opened 2023-12-04
Implements a version of tag generation where a common set of tags are repeatedly sampled

scottopell:
### What does this PR do?
When generating dogstatsd contexts, the current approach simply generates completely random tags until the number of desired contexts is reached, which works, but doesn't reflect real-world usage of dogstatsd.

In the real world, tags are repeated amongst messages frequently and its usually the unique combinations that create cardinality.

### Motivation

String interner experiments

### Related issues

This currently under-generates cardinality dramatically. If I request 100k contexts, I only get 1268 contexts. This is due to the way the common pool is sampled.

This algorithm needs some work, but posting this early as a draft in case anyone has feedback/ideas.

### Additional Notes


blt:
I like the direction this is taking.

scottopell:
closing in favor of #822

PR #765, opened 2023-12-05
Adjust shutdown signal to use an interior semaphore

blt:
### What does this PR do?

This commit adjusts our shutdown signal to make use of a semaphore. Now every party to the Shutdown will end up attempting to acquire on `sem` which has a capacity of MAX_PERMITS but has no permits until `signal` is called.

### Additional Notes

We guard against `Semaphore::add_permits` being filled over capacity: overfilling the semaphore will cause a panic. We might consider splitting `signal` off into its own type -- `Shutdown` becomes a parent with `signal` and a clonable child with `recv` -- but the existing situation is Good Enough we believe.

### Related issues

REF SMPTNG-104


lading/src/signals.rs
```diff
@@ -47,51 +36,45 @@ impl Shutdown {
...
         // If the shutdown signal has already been received, then return
         // immediately.
         if self.shutdown {
             return;
         }
 
-        // Cannot receive a "lag error" as only one value is ever sent.
-        let _ = self.notify.recv().await;
+        // We have no need of the permit that comes on the okay side, we also
+        // are fine to set shutdown if the semephore has been closed on us.
```
goxberry, line 58: semephore -> semaphore

lading/src/signals.rs
```diff
@@ -47,51 +37,47 @@ impl Shutdown {
...
+        let _ = self.sem.acquire().await;
```
goxberry, line 60: Do we want to drop this permit instead of forgetting it?

goxberry (approved):

lading/src/signals.rs
blt: This is a good question. Rust has a quirk where it treats `let _` as a special case assignment, dropping anything that's assigned as soon as the statement completes. If this had been:

```
let _res = self.sem.acquire().await;
```

I don't believe that same rule would apply. That is `_res` would be bound and in scope until the function exits, at which point Drop would be called.

lading/src/signals.rs
blt: Addressed in 317cc9a

lading/src/signals.rs
goxberry: I mean, would we want to do something like:

```rust
let permit = self.sem.acquire().await;
permit.forget(); // so permit is never released to semaphore
```

blt: Oh I see. I want this shutdown mechanism to be at least once so I think the current approach is acceptable, peers can receive the 'same' permit. If we wanted to ensure unique delivery forgetting the permit would be the way to go. 

There's an alternative interface here where `signal` is implemented on a different type from `recv`. I think if we want to go down that road we should carefully define semantics and maybe make a permit single-use but, as is, I'm content.

PR #771, opened 2023-12-08
AMLII-1328: Renames Shutdown to Phase

scottopell:
### What does this PR do?

`Shutdown` has been renamed to `Phase`

### Motivation
Will utilize this in a future PR to establish a "experimental phase has started" messaging mechanism utilizing `Phase`

### Related issues

for this Pull Request.

### Additional Notes

### DIFF
--- a/lading/src/target_metrics/prometheus.rs
+++ b/lading/src/target_metrics/prometheus.rs
@@ -11,7 +11,7 @@ use rustc_hash::FxHashMap;
 use serde::Deserialize;
 use tracing::{error, info, trace, warn};
 
-use crate::signals::Shutdown;
+use crate::signals::Phase;
 
 #[derive(Debug, Clone, Copy, thiserror::Error)]
 /// Errors produced by [`Prometheus`]
@@ -63,7 +63,7 @@ impl FromStr for MetricType {
 #[derive(Debug)]
 pub struct Prometheus {
     config: Config,
-    shutdown: Shutdown,
+    shutdown: Phase,
 }
 
 impl Prometheus {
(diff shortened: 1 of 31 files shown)


lading/src/target_metrics/prometheus.rs
```diff
@@ -91,16 +96,16 @@ impl Prometheus {
...
+            info!("Prometheus target metrics scraper running, but waiting for warmup to complete");
+            self.experiment_started.recv().await; // block until experimental phase entered
+            info!("Prometheus target metrics scraper starting collection");
             loop {
-                tokio::time::sleep(Duration::from_secs(1)).await;
-
                 let Ok(resp) = client
                     .get(&self.config.uri)
-                    .timeout(Duration::from_secs(1))
+                    .timeout(Duration::from_millis(800))
```
blt, line 108: I'm curious why we'd scrape at 200ms off a second. Intentional change here?

scottopell: yeah, a bit of a drive-by change, happy to revert from this PR and submit separately.

With the sleep and timeout both being 1s, my thought was that this could lead to the scraper falling behind, so I added 200ms of overhead.

goxberry: An ancillary issue with this sort of change is that the current iteration of regression detector analysis adopts the fiction that data points are spaced a second apart for simplicity so that we can use the `fetch_index` field of captures to compute rates. That said, now that our statistical analysis is in Rust, it would not be hard to change analysis to operate on timestamps -- it's changing roughly 10 lines in each of two places, and that change is one we were planning on making anyway[^1].

[^1]: An earlier implementation of statistical analysis in Rust used timestamps. See [`DataDog/single-machine-performance@ab023c1d`](https://github.com/DataDog/single-machine-performance/pull/1224/commits/ab023c1d1a6774b1778798472d398df8dbca943f) for the type of change that would need to be made in both the analysis command-line tool and in the analysis service.

scottopell: I'm splitting this up because I keep finding changes I want to make. This PR is now _just_ the rename of `Shutdown` -> `Phase`.

More will follow.

blt (approved):

PR #773, opened 2023-12-11
Hook procfs up into the generator, allow for end-user use

blt:
### What does this PR do?

This commit completes the work begun in #740 and exposes the purely faked procfs to end users. Nothing very clever is done in this PR. The only novelty is we avoid the use of a block cache and write stringy versions of the generated structs directly to disk. The 'server' constructed here idles entirely.

### Related issues

REF SMPTNG-88

### DIFF
--- a/lading/src/generator/procfs.rs
+++ b/lading/src/generator/procfs.rs
@@ -66,6 +66,10 @@ impl ProcFs {
     /// # Errors
     ///
     /// Returns an error if the config cannot be serialized.
+    ///
+    /// # Panics
+    ///
+    /// Function should never panic.
     pub fn new(config: &Config, shutdown: Shutdown) -> Result<Self, Error> {
         let mut rng = StdRng::from_seed(config.seed);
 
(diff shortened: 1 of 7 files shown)


GeorgeHahn (approved):

lading/src/generator/procfs.rs
```diff
@@ -23,6 +25,16 @@ pub enum Error {
...
     Io(#[from] ::std::io::Error),
+    /// Unable to strip prefix from passed copy file
+    #[error(transparent)]
+    Strip(#[from] StripPrefixError),
+}
+
+fn default_copy_from_host() -> Vec<PathBuf> {
+    vec![
+        PathBuf::from_str("/proc/uptime").unwrap(),
+        PathBuf::from_str("/proc/stat").unwrap(),
```
goxberry, line 36: As discussed in Slack, likely need to add `/proc/cpuinfo` due to `datadog-agent/pkg/metadata/host/host_notwin.go:67` calling https://github.com/shirou/gopsutil/blob/9deadc99147d80f732af3a59e624af73d0143891/cpu/cpu_linux.go#L176C51-L176C51.

PR #774, opened 2023-12-12
AMLII-1328: Adds an 'experimental' phase and removes warmup data from capture

scottopell:
### What does this PR do?

This PR implements:
1. An 'experimental' phase that will be entered after the `warmup_duration` expires
2. Modifies the target scrapers and capture manager to not begin execution until the experimental phase has been entered.
    i. As part of this, a test for `inner_main` was added which asserts that the expected # of `lading.running` entries given 
        a 1 second capture record interval.
4. Refactors the main `select!` in `lading/src/bin/lading.rs` to loop over the `select` to address a potential early-termination bug when the set of `generator`s are empty.
    a. This bug was encountered in https://github.com/DataDog/datadog-agent/pull/21452
6. Adds 2 gauge metrics:
    a. `lading.running` -- submits a value of `1` as long as the main `select!` of lading is executing
    b. `target.running` -- submits a value of `1` as long as the target is running.
7. `flush`es the capture file descriptor as soon as current lines are written, I observed in my unit tests that without this I could miss data. Likely due to the test runtime not being torn down by the time the capture files are inspected, but figure this data is valuable enough to pay the small IO penalty.

### Motivation
Especially visible in `smp local-run`, the generators and target scrapers have errors during the warmup phase connecting to the target. This is expected and is the entire point of the warmup phase.

This PR's original goal was simply to implement this "experimental phase", however scope creep got the best of me, which I hopefully made up for with my new unit tests for `inner_main`

### Additional Notes

There are a few `sleep` based loops that use a 1s timer, both the target metrics scraper, the target process observer, and the capture manager. Future improvement here would be to consider using `interval` instead of `sleep` in order to consider the [`MissedTickBehavior`](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html)

The new `gauge` metrics are useful to me in validating the resulting data in the capture file, however I can refactor to avoid these if desired.

### DIFF
--- a/lading/src/captures.rs
+++ b/lading/src/captures.rs
@@ -40,6 +40,7 @@ pub struct CaptureManager {
     capture_fp: BufWriter<std::fs::File>,
     capture_path: PathBuf,
     shutdown: Phase,
+    experiment_started: Phase,
     inner: Arc<Inner>,
     global_labels: FxHashMap<String, String>,
 }
@@ -50,7 +51,7 @@ impl CaptureManager {
     /// # Panics
     ///
     /// Function will panic if the underlying capture file cannot be opened.
-    pub async fn new(capture_path: PathBuf, shutdown: Phase) -> Self {
+    pub async fn new(capture_path: PathBuf, shutdown: Phase, experiment_started: Phase) -> Self {
         let fp = tokio::fs::File::create(&capture_path).await.unwrap();
         let fp = fp.into_std().await;
         Self {
@@ -59,6 +60,7 @@ impl CaptureManager {
             capture_fp: BufWriter::new(fp),
             capture_path,
             shutdown,
+            experiment_started,
             inner: Arc::new(Inner {
                 registry: Registry::atomic(),
             }),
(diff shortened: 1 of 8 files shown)


lading/src/captures.rs
```diff
@@ -50,7 +51,7 @@ impl CaptureManager {
     /// # Panics
     ///
     /// Function will panic if the underlying capture file cannot be opened.
-    pub async fn new(capture_path: PathBuf, shutdown: Phase) -> Self {
+    pub async fn new(capture_path: PathBuf, shutdown: Phase, experiment_started: Phase) -> Self {
```
GeorgeHahn, line 54: One cleanup would be to move the shutdown & started signals into a Signals struct. That would eliminate the risk of bugs where they're mixed up in a function call. Super optional.

GeorgeHahn (approved):
This looks great!

lading/src/captures.rs
scottopell: I like that thought, I'll file that away as a future improvement 👍

PR #775, opened 2023-12-13
Switches back to including warmup data in the capture manager

scottopell:
### What does this PR do?

Partially reverts #774 . The 'experimental' phase is still present, however the capture manager no longer waits for the experimental phase to be active before installing the metrics recorder.
This gives us warmup data in the capture files, which some later components rely on currently.

### Motivation

Some components rely on warmup data being present in the capture file.

### Related issues

#774


blt (approved):

GeorgeHahn (approved):
Thank you Scott! 🙏

PR #776, opened 2023-12-13
changelog: prep for 0.20.4 release

goxberry:
### What does this PR do?

Before cutting a `lading` release, its changelog must be updated. This commit updates the changelog in preparation for a 0.20.4 release.

### Motivation

Release changes that should fix broken `idle` experiment in `datadog-agent`.

### Related issues

n/a

### Additional Notes

n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-## [0.20.4]
+## [0.20.4-rc1]
 ### Changed
 - Dogstatsd payload structs now have public fields.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-## [0.20.4-rc1]
+## [0.20.4]
+### Added
+- `lading.running` gauge metric that submits a value of 1 as long as the main
+  `select!` of lading is executing.
+- `target.running` gauge metric that submits a value of 1 as long as the target
+  is running.
+
 ### Changed
 - Dogstatsd payload structs now have public fields.
+- Capture file descriptor is flushed as soon as current lines are written.
+- Renamed `Shutdown` data type to `Phase` so that this data type can be used as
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.20.4]
```
GeorgeHahn, line 9: Some PRs in this version are missing changelog entries; could you add those? Sorry, we're not too vigilant about that in PRs and don't currently have a CI check for it.

goxberry: Although the diff currently lists [five PRs](https://github.com/DataDog/lading/compare/v0.20.3...main) from the previous release, I count four, since https://github.com/DataDog/lading/pull/773 is included in the previous release, but due to the peculiar way we handle releases -- notably, releases are not necessarily commits in `main` -- that diff will list https://github.com/DataDog/lading/pull/773 as if it hadn't been included in the previous release.

https://github.com/DataDog/lading/pull/775 partially undoes https://github.com/DataDog/lading/pull/774, so to my mind, it's not worth calling out the parts of https://github.com/DataDog/lading/pull/775 and https://github.com/DataDog/lading/pull/774 that cancel out.

It seems like what might be missing is:

- Changes:
  - Capture file descriptor is flushed as soon as current lines are written.
- Fixes:
  - Refactors the main `select!` in `lading/src/bin/lading.rs` to loop over the select to address a potential early-termination bug when the set of generators are empty.
- Adds:
  - `lading.running` gauge metric -- submits a value of 1 as long as the main select! of lading is executing.
  - `target.running` gauge metric -- submits a value of 1 as long as the target is running.

Is that right?

scottopell: That looks correct to me, but just noting that the target scraper and prometheus http server (that exports the lading metrics) _do_ still respect the idea of the "experimental phase". Maybe worth to call out that we will have no target metrics during warmup.

(and sorry for continuing to forget to add changelog entries)

goxberry: Good call; thank you @GeorgeHahn and @scottopell for spotting the omissions.

Updated changelog in https://github.com/DataDog/lading/pull/776/commits/cc0817121f89c61569d8cc1edee765c31705fc0e.

blt (approved):

PR #787, opened 2024-01-25
Replace all Unwraps with Expect

dmalis18:
### What does this PR do?

This PR is the first step removing all unwraps and expects from lading. The goal of this PR is to first turn each unwrap into an expect so a user can at least get feedback when the program panics. The next step would be to completely remove expects by handling errors properly

### Motivation

In general it is not a good idea for the program to panic and it is much better stylistically if the program handles errors instead.

### Related issues

SMPTNG-114

### Additional Notes

### DIFF
--- a/integration/ducks/src/main.rs
+++ b/integration/ducks/src/main.rs
@@ -80,12 +80,12 @@ impl From<&HttpCounters> for HttpMetrics {
             median_entropy: val
                 .entropy
                 .quantile(0.5)
-                .expect("Error: quantile is outside of the valid range")
+                .expect("quantile is outside of the valid range")
                 .unwrap_or_default(),
             median_size: val
                 .body_size
                 .quantile(0.5)
-                .expect("Error: quantile is outside of the valid range")
+                .expect("quantile is outside of the valid range")
                 .unwrap_or_default(),
         }
     }
@@ -118,7 +118,7 @@ impl From<&SocketCounters> for SocketMetrics {
             median_entropy: val
                 .entropy
                 .quantile(0.5)
-                .expect("Error: quantile is outside of the valid range")
+                .expect("quantile is outside of the valid range")
                 .unwrap_or_default(),
         }
     }
@@ -130,9 +130,7 @@ async fn http_req_handler(req: Request<Body>) -> Result<hyper::Response<Body>, h
     let body = body::to_bytes(body).await?;
 
     {
-        let metric = HTTP_COUNTERS
-            .get()
-            .expect("Error: HTTP_COUNTERS not initialized");
+        let metric = HTTP_COUNTERS.get().expect("HTTP_COUNTERS not initialized");
         let mut m = metric.lock().await;
         m.request_count += 1;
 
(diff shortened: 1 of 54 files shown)


integration/ducks/src/main.rs
```diff
@@ -323,7 +341,9 @@ async fn main() -> Result<(), anyhow::Error> {
     debug!("hello from ducks");
 
     // Every ducks-sheepdog pair is connected by a unique socket file
-    let ducks_comm_file = std::env::args().nth(1).unwrap();
+    let ducks_comm_file = std::env::args()
+        .nth(1)
+        .expect("Error: The length of env args is less than 2");
```
goxberry, line 346: Could be worth rephrasing as "Error: ducks socket file argument missing" to provide more context.

lading/src/bin/lading.rs
```diff
@@ -80,13 +80,18 @@ impl FromStr for CliKeyValues {
         //
         // The approach taken here is to use the key notion as delimiter and
         // then tidy up afterward to find values.
-        static RE: Lazy<Regex> = Lazy::new(|| Regex::new(r"([[:alpha:]_]+)=").unwrap());
+        static RE: Lazy<Regex> = Lazy::new(|| {
+            Regex::new(r"([[:alpha:]_]+)=").expect("Error: Invalid regex pattern provided")
```
blt, line 84: I believe the "Error:" in these changes are redundant. The panic handler will signal something to the effect of a bad thing has happened.

blt (approved):
I'd prefer to remove the redundant `Error:` from the excepts here, but otherwise this looks good to me.

integration/ducks/src/main.rs
dmalis18: I agree. I will make that fix when I make the other changes.

blt (approved):

integration/ducks/src/main.rs
```diff
@@ -77,8 +77,16 @@ impl From<&HttpCounters> for HttpMetrics {
         HttpMetrics {
             request_count: val.request_count,
             total_bytes: val.total_bytes,
-            median_entropy: val.entropy.quantile(0.5).unwrap().unwrap_or_default(),
-            median_size: val.body_size.quantile(0.5).unwrap().unwrap_or_default(),
+            median_entropy: val
+                .entropy
+                .quantile(0.5)
+                .expect("Error: quantile is outside of the valid range")
```
goxberry: I'd suggest "Error: quantile argument must be between 0.0 and 1.0 inclusive", to provide more context.

goxberry (approved):

GeorgeHahn:
> 🚂 MergeQueue
> 
> This merge request is not mergeable yet, because of pending checks/missing approvals. It will be added to the queue as soon as checks pass and/or get approvals. **Note:** if you pushed new commits since the last approval, you may need additional approval. You can remove it from the waiting list with `/remove` command.
> 
> Use `/merge -c` to cancel this operation!

Huh, that's new to me. @dmalis18 we use the plain Github UI rather than devflow. (Either `Squash and merge` or `Enable auto-merge (squash)`)

It looks like devflow added a `devflow/merge` step that presumably isn't ever going to show ready. To merge this, you'll need to check `Merge without waiting for requirements to be met (bypass branch protections)` and then `Squash and merge`. We wouldn't do that normally because it means CI is still running or has failed.

PR #788, opened 2024-01-26
Linux 'observer' collects smaps entries and reports them into capture file

scottopell:
### What does this PR do?
Adds support in the linux observer to parse a given PIDs `/proc/pid/smaps` file in order to provide more detail about memory usage.

### Motivation
If RSS increases, it can be difficult to determine exactly _why_ rss went up.
Language-specific profilers and runtime telemetry will be valuable for the parts managed by that language's runtime, but OS-level details can be helpful as well.

### Related issues

REF [SMPTNG-185](https://datadoghq.atlassian.net/browse/SMPTNG-185)

### Additional Notes

[SMPTNG-185]: https://datadoghq.atlassian.net/browse/SMPTNG-185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lading/src/observer.rs
+++ b/lading/src/observer.rs
@@ -22,7 +22,6 @@ mod linux;
 mod memory;
 
 #[allow(dead_code)] // used on Linux
-#[allow(dead_code)] // used on Linux
 /// Expose the process' current RSS consumption, allowing abstractions to be
 /// built on top in the Target implementation.
 pub(crate) static RSS_BYTES: AtomicU64 = AtomicU64::new(0);
--- a/lading/src/observer.rs
+++ b/lading/src/observer.rs
@@ -18,6 +18,10 @@ use crate::signals::Phase;
 #[cfg(target_os = "linux")]
 mod linux;
 
+#[cfg(target_os = "linux")]
+mod memory;
+
+#[allow(dead_code)] // used on Linux
 #[allow(dead_code)] // used on Linux
 /// Expose the process' current RSS consumption, allowing abstractions to be
 /// built on top in the Target implementation.
--- a/lading/src/observer/memory.rs
+++ b/lading/src/observer/memory.rs
@@ -2,6 +2,7 @@ use regex::Regex;
 use std::{collections::HashMap, io::Read};
 
 const SMAP_SIZE_HINT: usize = 128 * 1024; // 128 kB
+const BYTES_PER_KIBIBYTE: u64 = 1024;
 
 #[derive(thiserror::Error, Debug)]
 /// Errors produced by functions in this module
@@ -33,7 +34,7 @@ pub(crate) struct Region {
     // empty string indicates no pathname
     pub(crate) pathname: String,
 
-    // Values (all in kibibytes)
+    // Values (all in bytes)
     pub(crate) size: u64,
     pub(crate) pss: u64,
     pub(crate) swap: u64,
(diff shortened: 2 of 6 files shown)


lading/src/observer.rs
```diff
@@ -18,6 +18,10 @@ use crate::signals::Phase;
 #[cfg(target_os = "linux")]
 mod linux;
 
+#[cfg(target_os = "linux")]
+mod memory;
+
+#[allow(dead_code)] // used on Linux
```
goxberry, line 24: Duplicate line?

lading/src/observer/memory.rs
```diff
@@ -0,0 +1,421 @@
...
+
+impl Region {
+    #[allow(clippy::similar_names)]
+    #[allow(clippy::too_many_lines)]
+    pub(crate) fn from_str(contents: &str) -> Result<Self, Error> {
+        let mut lines = contents.lines();
+
+        // parse metadata
+        // 7fffa9f39000-7fffa9f3b000 r-xp 00000000 00:00 0                          [vdso]
+        let metadata_line = lines.next().unwrap();
```
goxberry, line 61: Safety comment? Also not sure if we're continuing to avoid `unwrap` in favor or something like `expect`.

lading/src/observer/memory.rs
```diff
@@ -0,0 +1,421 @@
...
+        })
+    }
+}
+
+pub(crate) struct AggrMeasure {
+    pub(crate) size: u64,
+    pub(crate) pss: u64,
+    pub(crate) swap: u64,
+    pub(crate) rss: u64,
+}
```
goxberry, line 220: If I'm reading the tests correctly, these values have units of kibibytes[^1], but in other parts of this observer, similar quantities are tracked in units of bytes (e.g., `rss_bytes`). If we're intending to analyze this data, my preference would be to collect these statistics in bytes, even though there is potentially a false precision issue that would come into play (i.e., the data we're ingesting does not actually have precision to the nearest byte, but instead has precision to the nearest kibibyte).

[^1]: See https://github.com/torvalds/linux/blob/41bccc98fb7931d63d03f326a746ac4d429c1dd3/fs/proc/task_mmu.c#L808-L850, and note the right shift by 10 bits in the definition of the `SEQ_PUT_DEC` macro.

lading/src/observer/memory.rs
```diff
@@ -0,0 +1,421 @@
...
+    pub(crate) perms: String,
+    pub(crate) offset: u64,
+    // major:minor
+    pub(crate) dev: String,
+    // 0 indicates no inode
+    pub(crate) inode: u64,
+    // empty string indicates no pathname
+    pub(crate) pathname: String,
+
+    // Values (all in kB)
```
goxberry, line 36: All in kibibytes.

lading/src/observer/memory.rs
```diff
@@ -0,0 +1,421 @@
...
+                    let numeric = next_token(&mut chars).ok_or(Error::Parsing(format!(
+                        "Could not parse numeric value from line: {line}"
+                    )))?;
+                    swap = Some(numeric.parse::<u64>()?);
+                }
+                "Pss_Dirty:" => {
+                    let numeric = next_token(&mut chars).ok_or(Error::Parsing(format!(
+                        "Could not parse numeric value from line: {line}"
+                    )))?;
+                    pss_dirty = Some(numeric.parse::<u64>()?);
```
goxberry, line 184: I don't think the `Pss_Dirty` field exists in `/proc/{pid}/smaps` prior to Linux kernel version 6.0[^1]. Given that pre-6.0 LTS versions of the kernel are supported until December 2026[^2], it might make more sense for this field to be optional, if parsing it is necessary.

[^1]: Compare the [v6.0 version](https://github.com/torvalds/linux/blob/v6.0/fs/proc/task_mmu.c#L805-L846) to the [v5.19 version](https://github.com/torvalds/linux/blob/v5.19/fs/proc/task_mmu.c#L802-L842).
[^2]: According to https://www.kernel.org/category/releases.html, accessed 2024-01-29.

lading/src/observer/memory.rs
```diff
@@ -0,0 +1,421 @@
...
+
+        Ok(Regions(regions))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const TWO_REGION: &str = "
```
goxberry, line 288: Possibly worth noting that these outputs are representative of kernel versions 6.0 through 6.5, inclusive[^1].

[^1]: Because `Pss_Dirty` doesn't exist before version 6.0, as mentioned in https://github.com/DataDog/lading/pull/788/files#r1470328432, and a `KSM` field is added in version 6.6. See [v6.5 version](https://github.com/torvalds/linux/blob/v6.5/fs/proc/task_mmu.c) and [v6.6 version](https://github.com/torvalds/linux/blob/v6.6/fs/proc/task_mmu.c#L805-L847).

lading/src/observer/memory.rs
```diff
@@ -0,0 +1,534 @@
...
+                    let start = iter.peek()?.0;
+                    while let Some((_, c)) = iter.peek() {
+                        if c.is_whitespace() {
+                            break;
+                        }
+                        iter.next();
+                    }
+                    let end = iter.peek().map_or(line.len(), |&(idx, _)| idx);
+                    Some(&line[start..end])
+                };
```
goxberry, line 154: Minor, non-blocking nit: could be worth factoring this closure into a function because it's repeated earlier in the file.

goxberry (approved):

PR #791, opened 2024-01-31
SMPTNG-179: Replaces 'exe' with 'cmdline' for better process detection

scottopell:
### What does this PR do?

`exe` is not unique enough, this replaces it with `cmdline` which includes the `exe`.

### Motivation

colliding metrics when different logical processes run off same executable

### Additional Notes

Would appreciate any thoughts on what to do in the zombie case.

### DIFF
--- a/lading/src/observer/linux.rs
+++ b/lading/src/observer/linux.rs
@@ -84,6 +84,7 @@ impl Sampler {
         clippy::cast_possible_wrap
     )]
     pub(crate) fn sample(&mut self) -> Result<(), Error> {
+        // Key for this map is (pid, cmdline)
         let mut samples: FxHashMap<(i32, String), Sample> = FxHashMap::default();
 
         let mut total_processes: u64 = 0;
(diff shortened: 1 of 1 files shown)


lading/src/observer/linux.rs
```diff
@@ -154,25 +155,12 @@ impl Sampler {
...
-                    // clear _when_ this would be the case but, hey.
-                    String::new()
-                }
+            let cmdline: String = if let Ok(cmdline) = process.cmdline() {
+                cmdline.join(" ")
             } else {
-                continue;
+                // docs say that no cmdline means this is a zombie process
+                // what would be the most useful thing to do here?
+                String::new()
```
blt, line 163: Hmm. Maybe set the cmdline as "zombie"? I don't think we've ever seen one in practice, thb.

blt (approved):
I think temporarily keeping `exe` would be desirable just for gradual upgrades but otherwise I'm pro this change.

PR #798, opened 2024-02-01
Adjusts the 'capture' sleep to account for any time spent recording the captures

scottopell:
### What does this PR do?

Discount the amount of time slept in between capture intervals by the amount of time it took to record the captures.

### Motivation

Mostly preventative, want to avoid time drift in our capture records.

### DIFF
--- a/lading/src/captures.rs
+++ b/lading/src/captures.rs
@@ -13,7 +13,7 @@ use std::{
     io::{self, BufWriter, Write},
     path::PathBuf,
     sync::{atomic::Ordering, Arc},
-    time::{Duration, SystemTime, UNIX_EPOCH},
+    time::{Duration, Instant, SystemTime, UNIX_EPOCH},
 };
 
 use lading_capture::json;
@@ -190,7 +190,7 @@ impl CaptureManager {
         std::thread::Builder::new()
             .name("capture-manager".into())
             .spawn(move || loop {
-                let now = SystemTime::now();
+                let now = Instant::now();
                 if let Err(e) = self.record_captures() {
                     warn!(
                         "failed to record captures for idx {idx}: {e}",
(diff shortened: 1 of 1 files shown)


lading/src/captures.rs
```diff
@@ -193,7 +194,11 @@ impl CaptureManager {
                     info!("shutdown signal received");
                     return;
                 }
-                std::thread::sleep(Duration::from_secs(1));
+                // Sleep for 1 second minus however long we just spent recording captures
+                // assumption here is that the time spent recording captures is consistent
+                std::thread::sleep(
+                    Duration::from_secs(1) - now.elapsed().expect("Time went backwards"),
```
goxberry, line 200: Given @dmalis18's efforts to remove `expect` calls from Lading, it's probably best to add a variant to `captures::Error` covering this case, if possible, and use the question mark operator.

goxberry (approved):
Aside from the comment about the use of `expect`, this change looks good!

lading/src/captures.rs
goxberry: See https://github.com/DataDog/lading/pull/799/files#diff-a7d6b4f14763aee95b999b3f309f69f00a57760c25119ceb2a7f2a53333eeed4R38-R39.

blt: It's also not uncommon for time to go backward. We shouldn't crash here.

scottopell: @blt got any good reading for this? I'd assume that rust `SystemTime` is using a monotonic clock, which to my knowledge should prevent "time going backwards" bugs.

scottopell: >Distinct from the [Instant](https://doc.rust-lang.org/std/time/struct.Instant.html) type, this time measurement is not monotonic.
https://doc.rust-lang.org/std/time/struct.SystemTime.html

Well that's what I get for assuming! I'll switch this to use an `Instant` to record this duration I think

blt (approved):

PR #800, opened 2024-02-01
Moves smaps and smaps_rollup into an async task since these can be expensive

scottopell:
### What does this PR do?

The `sample` function of the `Observer` is now async and the `smaps` and `smaps_rollup` parsing has been moved into a spawned tokio task, which is `await`ed before the function returns.

### Motivation

I've heard that the `/proc/smaps` file can take up to 250ms to populate in some circumstances, attempt to offload this off the main `sample` thread for now.

### DIFF
--- a/lading/src/observer.rs
+++ b/lading/src/observer.rs
@@ -119,7 +119,7 @@ impl Server {
         loop {
             tokio::select! {
                 _ = sample_delay.tick() => {
-                    sampler.sample()?;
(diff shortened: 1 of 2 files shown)


blt (approved):

scottopell:
>        FAIL [   7.536s] lading::bin/lading tests::inner_main_capture_has_data
failure is known flake, going to override and merge here.

PR #802, opened 2024-02-02
Removing More Expects and Massive Refactor for Generator to remove expects in the future

dmalis18:
### What does this PR do?

This PR is for removing expects and replacing them with errors within Lading. The purpose is to clean up the overall quality of the code base using errors instead. It rehauls a lot of generator to give us the ability to do so more easily in the future

### Motivation

We want to make the program have handleable errors in comparison to crashing with panic.

### Related issues

SMPTNG-114

### Additional Notes

### DIFF
--- a/lading/src/generator/file_gen/logrotate.rs
+++ b/lading/src/generator/file_gen/logrotate.rs
@@ -57,6 +57,12 @@ pub enum Error {
     /// Failed to convert, value is 0
     #[error("Value provided must not be zero")]
     Zero,
+    /// Iterator has already finished
+    #[error("Next failed, iterator already finished")]
+    Next,
+    /// Empty block cache
+    #[error("Block cache does not have any blocks")]
+    EmptyBlockCache,
 }
 
 #[derive(Debug, Deserialize, PartialEq)]
@@ -303,12 +309,12 @@ impl Child {
         loop {
             // SAFETY: By construction the block cache will never be empty
             // except in the event of a catastrophic failure.
-            let blk = rcv.peek().await.expect("block cache is empty");
+            let blk = rcv.peek().await.ok_or(Error::EmptyBlockCache)?;
             let total_bytes = blk.total_bytes;
 
             tokio::select! {
                 _ = self.throttle.wait_for(total_bytes) => {
-                    let blk = rcv.next().await.expect("failed to advance through blocks"); // actually advance through the blocks
+                    let blk = rcv.next().await.ok_or(Error::Next)?; // actually advance through the blocks
                     let total_bytes = u64::from(total_bytes.get());
 
                     {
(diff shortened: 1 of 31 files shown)


blt:
> ### What does this PR do?
> A brief description of the change being made with this pull request.
> 
> ### Motivation
> What inspired you to submit this pull request?
> 
> ### Related issues
> A list of issues either fixed, containing architectural discussions, otherwise relevant for this Pull Request.
> 
> ### Additional Notes
> Anything else we should know when reviewing?

Do fill this out please. It's important for understanding the purpose of the PR and keeping track of what work it maps to.

lading/src/generator/splunk_hec.rs
```diff
@@ -118,6 +118,12 @@ pub enum Error {
     /// Failed to convert, value is 0
     #[error("Value provided must not be zero")]
     Zero,
+    /// Iterator has already finished
+    #[error("Next failed, iterator already finished")]
+    Next,
```
blt, line 123: This leaks internal implementation details and is not helpful to a user. Better to have something that is like "Attempted to iterate a closed iterator" if we want to keep the language specific to iterators. That said, I'm unsure that the _iterator_ is central to communicate here so much as we've run out of channels.

blt: We might also consider this an example of `except` not being all bad. Neither of these conditions should happen -- that is, we lack a SAFETY comment. Personally I'd vote to leave the excepts and not have the errors. They can't be handled anyway as these errors are catastrophic programming mistakes if they fire.

blt: Same comment for repeats of this pattern.

blt (changes requested):

lading/src/generator/splunk_hec.rs
dmalis18: That logic makes sense. I'll go back and reset them back to expects on Monday since it is more useful given the SAFETY issues if it fails.

lading/src/generator/file_gen/logrotate.rs
```diff
@@ -308,7 +308,7 @@ impl Child {
 
             tokio::select! {
                 _ = self.throttle.wait_for(total_bytes) => {
-                    let blk = rcv.next().await.expect("failed to advance through blocks"); // actually advance through the blocks
+                    let blk = rcv.next().await.expect("there is no next block in rcv"); // actually advance through the blocks
```
blt, line 311: I think the original language here was fine, but I'd be curious to know what you're attempting to communicate with the adjustment. My basic concern is "rcv" is the name of an internal variable and doesn't necessarily communicate anything to the end user that does not require internal knowledge of the program.

dmalis18: I did not realize that the rcv.nexts had different error messages when I originally made them to expect so I just took a standard message from one of them and placed it in all the spots that I had added the error types for it

dmalis18: Seems like a lot of these had unintentionally different expect messages

dmalis18: I am gonna back through and revert the changes to be more informative about the error and not about the internal details.

blt (approved):

PR #805, opened 2024-02-06
Removing More Expects from Lading

dmalis18:
### What does this PR do?

This PR is a continuation of the effort to remove expects from Lading and convert them into errors we can propagate. 

### Motivation

Cleans up the overall code quality we are working with

### Related issues

SMPTNG-114

### DIFF
--- a/lading/src/generator/unix_stream.rs
+++ b/lading/src/generator/unix_stream.rs
@@ -221,8 +221,7 @@ impl UnixStream {
                         let ready = stream
                             .ready(tokio::io::Interest::WRITABLE)
                             .await
-                            .map_err(Error::Io)
-                            .expect("Unable to start stream"); // Cannot ? in a spawned task :<. Mimics UDP generator.
+                            .map_err(Error::Io)?; // Cannot ? in a spawned task :<. Mimics UDP generator. Why does this compile????
                         if ready.is_writable() {
                             // Try to write data, this may still fail with `WouldBlock`
                             // if the readiness event is a false positive.
--- a/lading/src/generator/unix_stream.rs
+++ b/lading/src/generator/unix_stream.rs
@@ -221,7 +221,7 @@ impl UnixStream {
                         let ready = stream
                             .ready(tokio::io::Interest::WRITABLE)
                             .await
-                            .map_err(Error::Io)?; // Cannot ? in a spawned task :<. Mimics UDP generator. Why does this compile????
+                            .map_err(Error::Io)?;
                         if ready.is_writable() {
                             // Try to write data, this may still fail with `WouldBlock`
                             // if the readiness event is a false positive.
--- a/lading_payload/src/block.rs
+++ b/lading_payload/src/block.rs
@@ -33,6 +33,9 @@ pub enum SpinError {
     /// Error for crate deserialization
     #[error("Deserialization error: {0}")]
     Deserialize(#[from] crate::Error),
+    /// Error for constructing the block cache
+    #[error(transparent)]
+    ConstructBlockCache(#[from] ConstructBlockCacheError),
 }
 
 /// Error for [`Cache`]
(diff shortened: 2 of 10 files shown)


lading/src/generator/unix_stream.rs
```diff
@@ -221,8 +221,7 @@ impl UnixStream {
                         let ready = stream
                             .ready(tokio::io::Interest::WRITABLE)
                             .await
-                            .map_err(Error::Io)
-                            .expect("Unable to start stream"); // Cannot ? in a spawned task :<. Mimics UDP generator.
+                            .map_err(Error::Io)?; // Cannot ? in a spawned task :<. Mimics UDP generator. Why does this compile????
```
scottopell, line 224: I expect this is a stale comment, this is in a function that returns a `Result`, so `?` is valid here.

lading_payload/src/block.rs
```diff
@@ -711,15 +724,16 @@ where
...
     } else {
         let total_bytes = NonZeroU32::new(
             bytes
                 .len()
                 .try_into()
                 .expect("failed to get length of bytes"),
-        )?;
-        Some(Block { total_bytes, bytes })
+        )
+        .expect("total_bytes must be non-zero");
```
scottopell, line 735: an added `expect`? This should be `?`, no?

lading_payload/src/block.rs
```diff
@@ -53,6 +56,12 @@ pub enum Error {
     /// Error for crate deserialization
     #[error("Deserialization error: {0}")]
     Deserialize(#[from] crate::Error),
+    /// Error for constructing the block cache
+    #[error(transparent)]
+    ConstructBlockCache(ConstructBlockCacheError),
```
scottopell, line 61: duplicate `ConstructBlockCacheError` I believe, see  above `Construct` enum variant.

scottopell (approved):

lading_payload/src/block.rs
dmalis18: For some reason the option to use the ? operator went away after changing how Block worked and it was no longer working as it had prior. Basically introduced a compiler error somewhere.

blt: > Basically introduced a compiler error somewhere.

Specifically, the error message is this:

```
➜  lading git:(dmalis/remove_expect4) ✗ cargo check
    Checking lading-payload v0.1.0 (/Users/brian.troutwine/projects/com/github/datadog/lading/lading_payload)
error[E0277]: the `?` operator can only be used on `Result`s, not `Option`s, in a function that returns `Result`
   --> lading_payload/src/block.rs:734:10
    |
707 | / fn construct_block<R, S>(
708 | |     mut rng: &mut R,
709 | |     serializer: &S,
710 | |     chunk_size: u32,
...   |
713 | |     S: crate::Serialize,
714 | |     R: Rng + ?Sized,
    | |____________________- this function returns a `Result`
...
734 |           )?;
    |            ^ use `.ok_or(...)?` to provide an error compatible with `Result<std::option::Option<Block>, SpinError>`
    |
    = help: the trait `FromResidual<std::option::Option<Infallible>>` is not implemented for `Result<std::option::Option<Block>, SpinError>`
    = help: the following other types implement trait `FromResidual<R>`:
              <Result<T, F> as FromResidual<Yeet<E>>>
              <Result<T, F> as FromResidual<Result<Infallible, E>>>

For more information about this error, try `rustc --explain E0277`.
error: could not compile `lading-payload` (lib) due to previous error
```

The reason this happens is because the branch type above does not match, previously it was `Option<...>` but now it has to be `Result<Option<...>, ...>` on both branches. Additionally `?` will exit the function entirely and has to be the function's return type, so where previously it was an option now it needs to be a result option.

scottopell (approved):

PR #808, opened 2024-02-07
[SMPTNG-183] Compute and report working set for each process

paulcacheux:
### What does this PR do?

In the linux observer, in addition to the rss and similar memory metrics we already compute, now compute the working set of the matching cgroup. The value is computed using a similar heuristic as kubernetes.

### Motivation

Better understand of the impact of agent changes (especially related to the single agent binary project).

### Related issues

for this Pull Request.

### Additional Notes

### DIFF
--- a/lading/src/observer/linux.rs
+++ b/lading/src/observer/linux.rs
@@ -1,5 +1,6 @@
-use std::{collections::VecDeque, io, sync::atomic::Ordering};
+use std::{collections::VecDeque, io, path::Path, sync::atomic::Ordering};
 
+use cgroups_rs::cgroup::Cgroup;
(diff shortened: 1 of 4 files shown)


blt (approved):

lading/src/observer/linux.rs
```diff
@@ -350,6 +353,30 @@ impl Sampler {
                     gauge!("smaps_rollup.pss_shmem", v as f64, &labels);
                 }
             });
+
+            // if possible we compute the working set of the cgroup
```
scottopell, line 357: not sure how expensive this is, but if we think it may be somewhat expensive, lets put it in a `tokio::task` and spawn from the existing `JoinSet`

PR #821, opened 2024-02-21
Added a comment about cc behavior regarding the CFLAGS and the CXXFLAGS

dmalis18:
Added a comment about cc behavior regarding the CFLAGS and the CXXFLAGS

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -79,6 +79,7 @@ jobs:
         uses: taiki-e/upload-rust-binary-action@v1
 
         env:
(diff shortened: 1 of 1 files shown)


goxberry:
Small nit: could you edit the PR description so it reads "Added a comment about cc behavior regarding the CFLAGS and the CXXFLAGS"?

The description becomes the first non-header part of the PR's merge commit message, and those edits would make it easier to search the log for "comment", "CFLAGS", or "CXXFLAGS" in the event we need to search the git log for context.

goxberry (approved):
Approved with comment above re: PR description.

PR #822, opened 2024-02-27
SMPTNG-91: Improve dogstatsd tag generation to re-use existing tags

scottopell:
### What does this PR do?

This addresses https://datadoghq.atlassian.net/browse/SMPTNG-91

### Related issues
A previous attempt is at #761 

### Additional Notes
There was an attempt at creating a specialized "String Pool" I called `DelimitedPool` which would work by placing delimiters in the pre-generated `String` and then handing out `&str`s that contain one delimiter.  This would offer better efficiency, however proved tricky to sort out all the edge cases. The most recent version can be found [here](https://github.com/DataDog/lading/blob/a21985e53fad0488602881fdb745bb323da87e07/lading_payload/src/dogstatsd/common/delimitedpool.rs)

I think my latest attempt `of_newnew` is promising. I believe `of_newnew` would work if the pool generated based on resulting strings rather than where to place the delimiter. That is, I initially thought placing the delimiter every `desired_tag_length` spaces would be sufficient, but it is not compatible with the expanding delimiter approach taken in `of_newnew`.

I opted to revisit this later and instead used the existing `Pool`.


lading_payload/src/block.rs
```diff
@@ -120,6 +122,25 @@ pub fn default_cache_method() -> CacheMethod {
     CacheMethod::Fixed
 }
 
+fn write_blocks_to_disk(blocks: &[Block]) -> std::io::Result<PathBuf> {
```
scottopell, line 125: move this to another branch

lading_payload/benches/dogstatsd.rs
```diff
@@ -38,6 +38,6 @@ fn dogstatsd_all(c: &mut Criterion) {
 
 criterion_group!(
     name = benches;
-    config = Criterion::default().measurement_time(Duration::from_secs(90));
+    config = Criterion::default().measurement_time(Duration::from_secs(90)).sample_size(20);
```
scottopell, line 41: undo this

lading_payload/benches/default.rs
```diff
@@ -1,5 +1,6 @@
 use criterion::criterion_main;
 
+mod dogstatsd;
 mod tag;
 
-criterion_main!(tag::benches);
+criterion_main!(dogstatsd::benches);
```
scottopell, line 6: undo this

lading_payload/src/common/strings.rs
```diff
@@ -79,6 +82,27 @@ impl Pool {
         Some(&self.inner[lower_idx..upper_idx])
     }
 
+    /// Return a `&str` from the interior storage with size `bytes`. Result will
+    /// be `None` if the request cannot be satisfied.
+    pub fn of_size_with_handle<'a, R>(
```
scottopell, line 87: should I just update the regular `of_*` methods to include a handle?

lading_payload/src/common/strings.rs
```diff
@@ -21,6 +21,9 @@ pub struct Pool {
     inner: String,
 }
 
+/// Opaque 'str' handle for the pool
+pub type Handle = (usize, usize);
```
scottopell, line 25: handle is basically a weakptr

lading_payload/src/dogstatsd.rs
```diff
@@ -264,6 +261,13 @@ pub struct Config {
     /// a 4-byte header that is a little-endian u32 representing the
     /// total length of the data block.
     pub length_prefix_framed: bool,
+
+    /// The desired ratio between unique tags and the number of contexts
+    /// If this is 1, then that means every single tag is unique.
+    /// If this is 0, it is highly unlikely that the resulting traffic will
+    /// actually have a ratio of 0, the desired number of contexts will
+    /// _always_ be generated at the expense of this ratio.
+    pub tag_context_ratio: f32,
```
scottopell, line 270: update this name to match impl

lading_payload/src/dogstatsd.rs
```diff
@@ -279,8 +283,7 @@ impl Default for Config {
             },
             // https://docs.datadoghq.com/developers/guide/what-best-practices-are-recommended-for-naming-metrics-and-tags/#rules-and-best-practices-for-naming-metrics
             name_length: ConfRange::Inclusive { min: 1, max: 200 },
-            tag_key_length: ConfRange::Inclusive { min: 1, max: 100 },
-            tag_value_length: ConfRange::Inclusive { min: 1, max: 100 },
+            tag_length: ConfRange::Inclusive { min: 3, max: 100 },
```
scottopell, line 286: effective minimum is the same, 1 for key 1 for value and 1 for delimiter.

lading_payload/src/dogstatsd.rs
```diff
@@ -291,6 +294,7 @@ impl Default for Config {
             value: ValueConf::default(),
             // This should be enabled for UDS-streams, but not for UDS-datagram nor UDP
             length_prefix_framed: false,
+            tag_context_ratio: 0.2,
```
scottopell, line 297: default should be consistent with the `warn!` in the tag generator

lading_payload/src/dogstatsd.rs
```diff
@@ -460,14 +457,19 @@ impl MemberGenerator {
...
-            Rc::clone(&pool),
+            tag_length,
             num_contexts as usize,
-        );
+            tag_context_ratio,
+        ) {
+            Ok(tg) => tg,
+            Err(e) => {
+                warn!("Encountered error while constructing tag generator: {e}");
+                return Err(crate::Error::StringGenerate);
```
scottopell, line 470: ideally the underlying error can be surfaced in the actual error message, todo

lading_payload/src/dogstatsd/common.rs
```diff
@@ -12,6 +12,8 @@ use crate::{Error, Generator};
 use super::{ConfRange, ValueConf};
 
 pub mod tags;
+pub mod delimitedpool;
```
scottopell, line 15: rm this

lading_payload/src/dogstatsd/common/tags.rs
```diff
@@ -17,108 +21,333 @@ pub type Tagset = Vec<String>;
...
+/// 300 of these will result in new tags and 2700 of these will re-use an existing tag.
+/// With only 300 chances for new entropy, each individual tagset will likely over-sample
+/// the existing tags and therefore UNDER-generate the desired number of contexts.
 #[derive(Debug, Clone)]
 pub struct Generator {
     seed: Cell<u64>,
     internal_rng: RefCell<SmallRng>,
     tagsets_produced: Cell<usize>,
-    num_tagsets: usize,
+    desired_num_contexts: usize,
```
scottopell, line 47: hmm, should this be aware of "contexts"? Its not really necessary. I think we could revert this to `num_tagsets`? but I like the "desired" phrasing here, which implies that it may not be reached.

lading_payload/src/dogstatsd/common/tags.rs
```diff
@@ -17,108 +21,333 @@ pub type Tagset = Vec<String>;
...
-            num_tagsets,
+            desired_num_contexts,
+            unique_tag_probability,
+            unique_tags: RefCell::new(Vec::new()),
+        })
+    }
+
+    fn get_existing_tag<R>(
+        &self,
+        unique_tags: &RefMut<Vec<(Handle, Handle)>>,
```
scottopell, line 116: todo clean up the args here, I don't love this ref ref ref mut mut

lading_payload/src/dogstatsd/common/tags.rs
```diff
@@ -17,108 +21,333 @@ pub type Tagset = Vec<String>;
...
             .tags_per_msg
             .sample(&mut *self.internal_rng.borrow_mut()) as usize;
         for _ in 0..tags_count {
-            if self.tagsets_produced.get() >= self.num_tagsets {
-                // Reseed internal RNG with initial seed
-                self.internal_rng
-                    .replace(SmallRng::seed_from_u64(self.seed.get()));
-                self.tagsets_produced.set(0);
-            }
+            // if configured ratio is 1.0, then we always generate a brand-new tag
```
scottopell, line 165: comment is un-necessary now that its documented at component level

lading_payload/src/dogstatsd/common/tags.rs
```diff
@@ -17,108 +21,333 @@ pub type Tagset = Vec<String>;
...
     use rand::{rngs::SmallRng, SeedableRng};
 
-    use crate::dogstatsd::{strings, tags, ConfRange};
+    use crate::dogstatsd::{tags, ConfRange};
     use crate::Generator;
-    use std::rc::Rc;
+
+    /// given a list of tagsets, this returns the number of unique timeseries that they represent
+    /// in dogstatsd terms, if we assume a constant metric name,
+    fn count_num_contexts(tagsets: &Vec<tags::Tagset>) -> usize {
```
scottopell, line 230: should this itself have unit tests?

lading_payload/src/dogstatsd/common/tags.rs
scottopell: contexts don't _really_ make sense in terms of just tags. If the same tagset is applied to two different metric names, then that makes two contexts.

scottopell: it only makes sense in the usage of metric templates

lading_payload/src/dogstatsd.rs
scottopell: eh, this requires making `tags` public, which I don't care to do.

lading_payload/src/dogstatsd/common/tags.rs
scottopell: resolved, switched back to tagsets.

blt (approved):
Very nice.

PR #824, opened 2024-02-28
Adds generator that writes to a file on disk instead of socket/stream/etc

scottopell:
### What does this PR do?

This adds a new type of generator that takes the configured payload variant and writes the data out to a file on disk.
That's it.

### Motivation

Helpful to see the final payload output without having to pipe through `socat` or similar.

### DIFF
--- a/lading/src/generator/passthru_file.rs
+++ b/lading/src/generator/passthru_file.rs
@@ -150,6 +150,7 @@ impl PassthruFile {
     ///
     /// # Errors
     ///
+    /// If the underlying block cache fails to spin, an error will be returned.
     ///
     /// # Panics
     ///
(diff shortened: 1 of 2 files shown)


lading/src/generator/passthru_file.rs
```diff
@@ -0,0 +1,217 @@
...
+            metric_labels: labels,
+            shutdown,
+        })
+    }
+
+    /// Run [`PassthruFile`] to completion or until a shutdown signal is received.
+    ///
+    /// # Errors
+    ///
+    ///
```
goxberry, line 153: Needs text for the "Errors" section.

blt (approved):

blt (approved):

PR #829, opened 2024-02-29
SMPTNG-259: Improve block chunk generation by re-using block sizes freely

scottopell:
### What does this PR do?

When generating the block chunks that will be filled with a payload, instead of attempting to maintain even weightings of the block sizes, freely re-use them as needed to provide more capacity in the block chunks.

This also provides useful warnings when the payload cannot be fulfilled as written, eg:
```
2024-02-29T22:21:42.155929Z  WARN lading_payload::block: Failed to construct chunks adding up to 100 MB. Chunks created have total capacity of 10.24 MB. 89.76 MB unfulfilled. Max capacity of chunks was hit, consider making block sizes larger to fit more data.
```

### Motivation

User asked for a certain amount of data, lets try harder to fulfill their request.

### Additional Notes
Note currently based on #828

### DIFF
--- a/lading_payload/src/block.rs
+++ b/lading_payload/src/block.rs
@@ -532,14 +532,15 @@ fn chunk_bytes<const N: usize>(
         }
     }
 
+    let computed_chunk_capacity = total_bytes.get() - bytes_remaining;
+    let total_chunk_capacity = byte_unit::Byte::from_u64(computed_chunk_capacity.into());
+    let total_chunk_capacity_str = total_chunk_capacity
+        .get_appropriate_unit(byte_unit::UnitType::Decimal)
+        .to_string();
+
     if bytes_remaining > 0 {
         let total_requested_bytes = byte_unit::Byte::from_u64(total_bytes.get().into());
         let total_requested_bytes_str = total_requested_bytes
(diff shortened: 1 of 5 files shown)


lading_payload/src/block.rs
```diff
@@ -729,6 +746,61 @@ mod test {
...
+        assert_eq!(total_bytes.get(), block_chunk_sum);
+    }
+
+    proptest! {
+        // This test does not pass! (hence the #[ignore])
+        // It is trivial to find a combination of total_bytes and `block_byte_sizes` where
+        // its not possible to reach the max amount. Consider total_bytes=5 and block_sizes=[4]
+
+        // proptest todo - figure out how to express "block_byte_sizes members must be less than total_bytes"
+        // until then, the strategy is to set the max block size to the min total_bytes
```
blt, line 789: You could write a strategy for this. https://proptest-rs.github.io/proptest/proptest/tutorial/higher-order.html. There's also https://docs.rs/proptest/latest/proptest/macro.prop_assume.html but that'll throw away a ton of cases, where if you build values that are less than `total_bytes` and glob them into a vec you're good.

blt (approved):

PR #835, opened 2024-03-05
De-duplicate block sizes

scottopell:
### What does this PR do?

Refactors the generators that take in `block_sizes` as an argument and extracts that logic and default to a single location in the code.

### Motivation

Simplify my next set of changes and de-duplicate the code.

### DIFF
--- a/lading/src/generator/udp.rs
+++ b/lading/src/generator/udp.rs
@@ -18,7 +18,7 @@ use std::{
     time::Duration,
 };
 
-use byte_unit::{Byte, ByteError, ByteUnit};
+use byte_unit::ByteError;
 use lading_throttle::Throttle;
 use metrics::{counter, gauge, register_counter};
 use rand::{rngs::StdRng, SeedableRng};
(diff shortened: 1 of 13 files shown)


lading/src/generator/udp.rs
```diff
@@ -95,27 +95,8 @@ impl Udp {
...
-                    Byte::from_unit(4_f64, ByteUnit::MB),
-                ],
-                |sizes| sizes.iter().map(|sz| Ok(*sz)).collect::<Vec<_>>(),
-            )
-            .into_iter()
-            .collect::<Result<Vec<_>, _>>()?
-            .iter()
-            .map(|sz| NonZeroU32::new(sz.get_bytes() as u32).expect("bytes must be non-zero"))
-            .collect();
+        // TODO pass in datagram_friendly: true
```
blt, line 98: Left a TODO here.

lading_payload/src/block.rs
```diff
@@ -726,6 +722,58 @@ where
...
+    ]
+    .iter()
+    .map(|sz| {
+        NonZeroU32::new(
+            u32::try_from(sz.get_bytes()).expect("Block size not representable as 32bit integer"),
+        )
+        .expect("Block size is non-zero")
+    })
+    .collect()
+}
```
blt, line 775: This is so much nicer than having this copied everywhere.

blt (approved):

lading/src/generator/udp.rs
scottopell: intentional, this will be the next PR

PR #842, opened 2024-03-07
Improve lading container builds

blt:
### What does this PR do?

This commit attempts to build lading on amd64 and arm64. I will also attempt improvements in build speed. Note we now only build the lading binary when building containers.

### Related issues

REF SMPTNG-151

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -18,23 +18,23 @@ jobs:
       packages: write
 
     steps:
-      - uses: actions/checkout@v2.3.5
+      - uses: actions/checkout@v4
         with:
           sparse-checkout: .
 
       - name: Set up Docker Buildx
         id: buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v3
 
       - name: Log in to the Container registry
-        uses: docker/login-action@v1
+        uses: docker/login-action@v3
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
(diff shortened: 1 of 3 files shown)


GeorgeHahn (approved):

.github/workflows/container.yml
```diff
@@ -8,7 +8,54 @@ env:
...
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        platform: [linux/arm64]
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - uses: actions/checkout@v2.3.5
```
scottopell, line 21: I noticed in another container build job that the git checkout is fetching all branches and tags.
I believe we could get by with a sparse checkout https://github.com/actions/checkout?tab=readme-ov-file#fetch-only-the-root-files

blt: Oh I bet you're right! I'll fix that here.

blt: Addressed in [a84f3fd](https://github.com/DataDog/lading/pull/842/commits/a84f3fde270b69b5e2e132839429e760c76be648).

scottopell: Not quite, gotta bump the version too , `sparse-checkout` didn't exist in v2

blt: Addressed in [1683c68](https://github.com/DataDog/lading/pull/842/commits/1683c68c5b8ebb8041013c53d71dba7fe285480f) I do believe.

PR #852, opened 2024-03-12
Attempt to fix cargo-deny CI failure

scottopell:
### What does this PR do?

Removes some entries from `deny.toml`

### Motivation

Fix CI for `cargo-deny`

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
 channel = "1.75.0"
-profile = "complete"
+profile = "minimal"
--- a/deny.toml
+++ b/deny.toml
@@ -14,11 +14,21 @@ allow = [
 ]
 unused-allowed-license = "allow"
 
+copyleft = "deny"
+default = "deny"
+
 [sources]
 unknown-git = "deny"
 
 [advisories]
-ignore = [ ]
+vulnerability = "deny"
+unmaintained = "deny"
+notice = "deny"
+unsound = "deny"
+ignore = [
+  # atty is unsound for some custom allocators on Windows and is unmaintained
+  "RUSTSEC-2021-0145",
+]
 
 [bans]
 multiple-versions = "allow"
--- a/deny.toml
+++ b/deny.toml
@@ -14,21 +14,11 @@ allow = [
 ]
 unused-allowed-license = "allow"
 
-copyleft = "deny"
-default = "deny"
-
 [sources]
 unknown-git = "deny"
 
 [advisories]
-vulnerability = "deny"
-unmaintained = "deny"
-notice = "deny"
-unsound = "deny"
-ignore = [
-  # atty is unsound for some custom allocators on Windows and is unmaintained
-  "RUSTSEC-2021-0145",
-]
+ignore = [ ]
 
 [bans]
 multiple-versions = "allow"


rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
 channel = "1.75.0"
-profile = "complete"
+profile = "minimal"
```
goxberry, line 3: I think it'd be worth adding component overloads for the tools we use in CI via

```toml
components = [ "rustfmt", "clippy" ]
```

https://github.com/DataDog/lading/commit/7cb4ee80844d1b53b3de6a310df2c6ecde5897fa mentions a desire to install `rust-analyzer` by default, so `"rust-analyzer"` could be worth adding to the `components` array as well.

scottopell: Do you know what this affects? I'm not totally clear on what changing this does in a normal developer workflow.

for CI workflows, it seems like each job/action works in its own way as far as "what rust components do I need". For `fmt/clippy/check`, I handle these explicitly https://github.com/DataDog/lading/blob/main/.github/workflows/ci.yml#L24-L34

And the `cargo-deny` action is separate and apparently does look at the `rust-toolchain.toml`, whereas `actions-rust-lang/setup-rust-toolchain@v1.8` does not appear to look at this file.

blt (approved):

rust-toolchain.toml
scottopell: merging this now to fix CI, but happy to make a follow up to address this issue.

goxberry: > And the `cargo-deny` action is separate and apparently does look at the `rust-toolchain.toml`, whereas `actions-rust-lang/setup-rust-toolchain@v1.8` does not appear to look at this file.

Quoting the [`actions-rust-lang/setup-rust-toolchain` README](https://github.com/actions-rust-lang/setup-rust-toolchain/blob/main/README.md#inputs):

> If a [toolchain file](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) (i.e., `rust-toolchain` or `rust-toolchain.toml`) is found in the root of the repository and no `toolchain` value is provided, all items specified in the toolchain file will be installed. If a `toolchain` value is provided, the toolchain file will be ignored. If no `toolchain` value or toolchain file is present, it will default to `stable`.

As of commit https://github.com/DataDog/lading/commit/80475de0b96d8b1e57aa9100baf5b8a20f8acb27, there are invocations of `actions-rust-lang/setup-rust-toolchain` in 4 jobs of `.github/workflows/ci.yml`. None of these invocations specifies a `toolchain` value, including the invocation linked in https://github.com/DataDog/lading/blob/main/.github/workflows/ci.yml#L24-L34, so all four invocations use the `rust-toolchain.toml` file, and the specified `components` values in that link install component overrides to the toolchain specified in `rust-toolchain.toml`. Specifying components in `rust-toolchain.toml` would make the `components` values in `.github/workflows/ci.yml` unnecessary.

The `rust-toolchain.toml` file also affects local builds of the project[^2], and will install the specified toolchain profile and components if absent. The `"complete"` profile installs all available components, and thus should have installed `rustfmt`, `clippy`, and so on already, so reverting the profile to `"minimal"` will have no apparent effect for a developer that has already build `lading` using a `1.75.0` toolchain with the `"complete"` profile.

For some reason, specifying the `"complete"` profile will cause `actions-rust-lang/setup-rust-toolchain` to attempt to install the `miri` component. This component is only available in nightly builds of Rust[^1], so attempting to install it in a `1.75.0` toolchain should cause an error. This error was causing that action to crash when we invoked `actions-rust-lang/setup-rust-toolchain` with the `rust-toolchain.toml` file in commit https://github.com/DataDog/lading/commit/36d650b82c4779027e474540690b1313e410fada. This error was also the motivation for submitting https://github.com/DataDog/single-machine-performance/pull/1726 to `single-machine-performance`. This error does not occur when building the project locally, and it's unclear to me why there is a difference in behavior for the `"complete"` profile in CI, but I don't think it's worth trying to figure out the source of that difference.

[^1]: See, for instance, the discussion in the ["Miri" section of Learning Rust with Entirely Too Many Linked Lists](https://rust-unofficial.github.io/too-many-lists/fifth-miri.html).
[^2]: See the ["Overrides" chapter of the `rustup` book](https://rust-lang.github.io/rustup/overrides.html).

PR #855, opened 2024-03-12
Add Scott to the authors list, fix a syntax issue with George

blt:
### What does this PR do?

Fair to say that Scott's got as much to do with lading's internals as me and George at this point. Adding him to the authors list.


scottopell (approved):
🎉

PR #856, opened 2024-03-13
Fix block tearing during datagram generation

scottopell:
### What does this PR do?

Adds a new concept to the `Block` that describes valid strategies for splitting up the block into smaller chunks.
A lot of payloads are newline-delimited, so that's what has been implemented.

### Motivation

Fixing a bug in the unix_datagram generator where invalid messages will get sent through.

### Additional Notes

I don't know how much I like this approach. While its more correct, it essentially converts the chunked data back into a stream and then creates tons of tiny datagrams.

We don't have a good story around "how do I control the datagram sizes", so it may be that this is better than before and we should invest in coming up with a better story for datagram sizing.


scottopell:
this approach wasn't very good. #879 is the better solution

PR #861, opened 2024-03-19
block: add failing case to proptest cases

goxberry:
### What does this PR do?

This commit adds a failing case to the `proptest` cases for `block::test::construct_block_cache_inner_fills_all_blocks` in `lading_payload`.

### Motivation

Test failure occurred in testing of #847; see https://github.com/DataDog/lading/actions/runs/8336453710/job/22813606919?pr=847.

### Related issues

See above.

### Additional Notes

n/a

PR #862, opened 2024-03-19
Fix for PropTest failure in block construction

scottopell:
### What does this PR do?

Provides a fix for the issue that proptest found in https://github.com/DataDog/lading/pull/861

### Motivation

The issue here lies with the fact that this test is trying to fill a 100byte block using the `json` generator, however the json generator randomly generates messages, so if they are too big, it simply tries again.
The threshold for "trying again" is quite low and therefore this issue is hit regularly with a 100byte block.

Increasing the block size to 1000 preserves the intent of the test while recognizing that small blocks are hard to generate JSON for.

### DIFF
--- a/lading_payload/src/block.rs
+++ b/lading_payload/src/block.rs
@@ -831,7 +831,7 @@ mod test {
             use rand::{rngs::SmallRng, SeedableRng};
 
             let mut rng = SmallRng::seed_from_u64(seed);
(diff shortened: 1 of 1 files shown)


goxberry:
This PR doesn't need a changelog entry, right?

goxberry (approved):

PR #870, opened 2024-04-02
Capture manager should check shutdown before attempting to record captures

scottopell:
### What does this PR do?

Addresses a shutdown race that can occur where `record_captures` is called incorrectly after the experiment has shut down.
This can result in truncated `capture.json` result files.

### Motivation

Fixing a bug in capture writing.

### DIFF
--- a/lading/src/captures.rs
+++ b/lading/src/captures.rs
@@ -194,6 +194,7 @@ impl CaptureManager {
             .name("capture-manager".into())
             .spawn(move || loop {
                 if self.shutdown.try_recv() {
+                    // TODO - see SMPTNG-340 for details about more synchronization needed here
                     info!("shutdown signal received");
                     return;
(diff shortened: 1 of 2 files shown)


lading/src/captures.rs
```diff
@@ -193,6 +193,10 @@ impl CaptureManager {
         std::thread::Builder::new()
             .name("capture-manager".into())
             .spawn(move || loop {
+                if self.shutdown.try_recv() {
+                    info!("shutdown signal received");
+                    return;
+                }
```
GeorgeHahn, line 199: Do we need to wait in the main thread until this is hit?

GeorgeHahn: @scottopell mind linking that ticket here?

GeorgeHahn (approved):

PR #881, opened 2024-04-25
procfs generator: wrap `comm` in parens

GeorgeHahn:
### What does this PR do?

Bugfix for the procfs generator - the `comm` field should be wrapped in parenthesis.

From the manpage:

/proc/pid/stat
...
(2) comm  %s
The filename of the executable, in parentheses. Strings longer than TASK_COMM_LEN (16) characters (including the terminating null byte) are silently truncated.  This is visible whether or not the executable is swapped out.

### Related issues

https://github.com/DataDog/smp-dogfooding/pull/13#issuecomment-2025938017

### Additional notes

It's unclear how the process agent experiments have functioned at all - this bug will reliably crash the gopsutil code. It seems likely there's a race condition or some unexpected logic that allows process agent experiments to pass enough of the time for jobs to pass. (It could also indicate something incorrect in the SMP system.)

### DIFF
--- a/lading_payload/src/procfs.rs
+++ b/lading_payload/src/procfs.rs
@@ -1683,7 +1683,7 @@ impl<'a> Generator<'a> for ProcessGenerator {
     {
         // For simplicity, assume the command line is a single string consisting
         // of a single path component and no arguments.
-        let cmdline_size: usize = rng.gen_range(1..NAME_MAX); // is this range off by one?
+        let cmdline_size: usize = rng.gen_range(1..=NAME_MAX);
 
         // SAFETY: If this call fails, then execution should panic because an
         // inability to generate process command lines is a serious bug.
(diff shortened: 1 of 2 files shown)


blt (approved):

lading_payload/src/procfs.rs
```diff
@@ -1683,7 +1683,7 @@ impl<'a> Generator<'a> for ProcessGenerator {
     {
         // For simplicity, assume the command line is a single string consisting
         // of a single path component and no arguments.
-        let cmdline_size: usize = rng.gen_range(1..NAME_MAX);
+        let cmdline_size: usize = rng.gen_range(1..NAME_MAX); // is this range off by one?
```
goxberry, line 1686: Yes, that range is off by one because Rust's strings are not null-terminated, and that distinction slipped my mind. Good catch!

goxberry (approved):

lading_payload/src/procfs.rs
GeorgeHahn: Whoops! I accidentally committed this comment. Thanks for checking, I'll replace this comment with a fix.

PR #882, opened 2024-04-25
Allow peers to signal Phase hold

blt:
### What does this PR do?

This commit allows peers of the `Phase` to signal to the signaler that they have not yet completed their work. We achieve this by keeping a tally of the total number of peers that have called `register` -- a new function -- and require that each peer maintain a `Token` which on drop adds permits to a new semaphore. On the signaler side if the number of peers equals the number of permits in the ack'ing semaphore all registered peers have exited.

### Related issues

REF SMPTNG-356

### DIFF
--- a/lading/src/captures.rs
+++ b/lading/src/captures.rs
@@ -192,28 +192,32 @@ impl CaptureManager {
 
         std::thread::Builder::new()
             .name("capture-manager".into())
-            .spawn(move || loop {
-                if self.shutdown.try_recv() {
-                    // TODO - see SMPTNG-340 for details about more synchronization needed here
-                    info!("shutdown signal received");
-                    return;
+            .spawn(move || {
+                let token = self.shutdown.register();
+                loop {
+                    if self.shutdown.try_recv() {
+                        // TODO - see SMPTNG-340 for details about more synchronization needed here
+                        info!("shutdown signal received");
+                        drop(token);
+                        return;
+                    }
+                    let now = Instant::now();
+                    if let Err(e) = self.record_captures() {
+                        warn!(
+                            "failed to record captures for idx {idx}: {e}",
+                            idx = self.fetch_index
+                        );
+                    }
+                    self.fetch_index += 1;
+                    // Sleep for 1 second minus however long we just spent recording captures
+                    // assumption here is that the time spent recording captures is consistent
+                    let delta = now.elapsed();
+                    if delta > Duration::from_secs(1) {
+                        warn!("Recording capture took more than 1s (took {delta:?})");
+                        continue;
+                    }
+                    std::thread::sleep(Duration::from_secs(1).saturating_sub(delta));
                 }
-                let now = Instant::now();
-                if let Err(e) = self.record_captures() {
-                    warn!(
-                        "failed to record captures for idx {idx}: {e}",
-                        idx = self.fetch_index
-                    );
-                }
-                self.fetch_index += 1;
-                // Sleep for 1 second minus however long we just spent recording captures
-                // assumption here is that the time spent recording captures is consistent
-                let delta = now.elapsed();
-                if delta > Duration::from_secs(1) {
-                    warn!("Recording capture took more than 1s (took {delta:?})");
-                    continue;
-                }
-                std::thread::sleep(Duration::from_secs(1).saturating_sub(delta));
             })?;
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


lading/src/captures.rs
```diff
@@ -192,28 +192,32 @@ impl CaptureManager {
...
-            .spawn(move || loop {
-                if self.shutdown.try_recv() {
-                    // TODO - see SMPTNG-340 for details about more synchronization needed here
-                    info!("shutdown signal received");
-                    return;
+            .spawn(move || {
+                let token = self.shutdown.register();
+                loop {
+                    if self.shutdown.try_recv() {
+                        // TODO - see SMPTNG-340 for details about more synchronization needed here
```
scottopell, line 199: stale comment at this point

lading/src/signals.rs
```diff
@@ -19,10 +22,16 @@ use tracing::info;
...
-    /// The interior semaphore, the mechanism by which we will 'broadcast'
-    /// the current phase has been entered.
+    /// The mechanism by which we will 'broadcast' the current phase has been
+    /// entered.
     sem: Arc<Semaphore>,
 
+    /// The mechanism by which we will 'ack' receipt of phase change to the `Phase` creator.
+    ack_sem: Arc<Semaphore>,
+
+    /// The total number of peers, incremented only by `clone`.
```
scottopell, line 32:
```suggestion
    /// The total number of peers, incremented only by `register`.
```

lading/src/signals.rs
```diff
@@ -89,13 +100,50 @@ impl Phase {
...
     }
+
+    /// Wait for any registered peers before returning.
+    #[tracing::instrument]
+    pub async fn wait_for_peers(self) {
+        let peers = self.peers.load(Ordering::Acquire);
+        let _ = self.ack_sem.acquire_many(peers).await;
+    }
+
+    /// Register with the `Phase` owner to avoid the call to `signal` from
```
scottopell, line 111:
```suggestion
    /// Register with the `Phase` creator to avoid the call to `signal` from
```
?

not sure, the word "owner" just feels a bit nebulous here given the `clone` idiom.

lading/src/signals.rs
```diff
@@ -89,13 +100,50 @@ impl Phase {
...
+    /// proceeding without the token returned here being dropped.
+    #[tracing::instrument]
+    pub fn register(&self) -> Token {
+        // Increment the peers. We are careful to to AcqRel this fetch and store
```
scottopell, line 115:
```suggestion
        // Increment the peers. We are careful to AcqRel this fetch and store
```

lading/src/signals.rs
```diff
@@ -89,13 +100,50 @@ impl Phase {
...
+        // to avoid the parent from being unable to read the correct number of
```
scottopell, line 116: What is the scenario where this could happen? For those of us less well-versed in memory orderings, this is not obvious.

scottopell (commented):
looks good to me! Some comments on phrasing, but I like the `register`/`drop` idiom here, it feels pretty intuitive.

Since this component is now pretty complex, can you add an example usage and maybe some tests if they seem useful?

lading/src/signals.rs
blt: If you did this with relaxed memory ordering the count could be wrong, probably on the low side but not necessarily. On the low side you exit the process before all the registered peers are wrapped up, on the high side you never exit. Acquire / Release semantics means that the reads writes end up ordered properly so we have an accurate count. 

If it's more clear we might choose to use a Mutex here. Increasingly I think that this struct should be split between the creator and the peers and if we do that it'd be the appropriate time to put a mutex in.

lading/src/captures.rs
blt: Addressed in [9f36432](https://github.com/DataDog/lading/pull/882/commits/9f364320a7fd68c4babd867119e00eeb030cbc71)

lading/src/signals.rs
scottopell: > Increasingly I think that this struct should be split between the creator and the peers and if we do that it'd be the appropriate time to put a mutex in.

I can't see the specific structure, but I do like the "creator" and "user" being more separate, its a bit confusing that the "users" just operate on cloned instances

scottopell (approved):
Looks good, would still like to see a test or two, but fine as-is as well

PR #883, opened 2024-04-26
Integration test fixes

GeorgeHahn:
### What does this PR do?

Fix integration test issues seen in https://github.com/DataDog/lading/pull/877

### DIFF
--- a/integration/sheepdog/src/lib.rs
+++ b/integration/sheepdog/src/lib.rs
@@ -19,7 +19,12 @@
 //! `RUST_LOG=trace cargo test -p sheepdog`
 //!
 
-use std::{io::Write, path::PathBuf, process::Stdio, time::Duration};
+use std::{
+    io::{Read, Write},
+    path::PathBuf,
+    process::Stdio,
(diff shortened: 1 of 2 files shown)


integration/sheepdog/src/lib.rs
```diff
@@ -70,14 +81,43 @@ pub fn build_lading() -> Result<PathBuf, anyhow::Error> {
...
+        match self {
+            TakeableTempDir::Owned(inner) => {
+                let dir = inner.path().to_owned();
+                let tempdir = std::mem::replace(self, TakeableTempDir::Borrowed(dir));
+                Ok(tempdir)
+            }
+            TakeableTempDir::Borrowed(_) => anyhow::bail!("cannot take a borrowed tempdir"),
+        }
+    }
+}
```
goxberry, line 111: If I'm understanding this code correctly, this construction is a really interesting way to consume a field of a type that does not implement `Clone` without dropping the instance contained in that field.

goxberry (approved):

PR #886, opened 2024-05-01
Allow for dynamic block sizes in block construction

blt:
### What does this PR do?

This commit improves the 'pack' of our block construction by allow for dynamic block sizes based around the user-provided block sizes. The key insight here is that users don't _really_ care about the block sizes and in practice they are not obeyed: serialized blocks vary depending on the format. They're an abstraction leak. We therefore dynamically build blocks, keeping track of the minimum floor for a given serialization format and searching both 20% more and less on either size of the user-configured blocks.

### DIFF
--- a/lading_payload/src/block.rs
+++ b/lading_payload/src/block.rs
@@ -183,7 +183,7 @@ impl Cache {
                 );
                 let _guard = span.enter();
 
-                construct_block_cache_inner(&mut rng, &ta, &block_chunks)?
+                construct_block_cache_inner(&mut rng, &ta, &block_chunks, total_bytes.get())?
             }
             crate::Config::Syslog5424 => {
                 let span = span!(
@@ -194,7 +194,12 @@ impl Cache {
                 );
                 let _guard = span.enter();
 
-                construct_block_cache_inner(&mut rng, &crate::Syslog5424::default(), &block_chunks)?
+                construct_block_cache_inner(
+                    &mut rng,
+                    &crate::Syslog5424::default(),
+                    &block_chunks,
+                    total_bytes.get(),
+                )?
             }
             crate::Config::DogStatsD(conf) => {
                 match conf.valid() {
@@ -214,7 +219,12 @@ impl Cache {
                 );
                 let _guard = span.enter();
 
-                construct_block_cache_inner(&mut rng, &serializer, &block_chunks)?
+                construct_block_cache_inner(
+                    &mut rng,
+                    &serializer,
+                    &block_chunks,
+                    total_bytes.get(),
+                )?
             }
             crate::Config::Fluent => {
                 let pyld = crate::Fluent::new(&mut rng);
@@ -225,7 +235,7 @@ impl Cache {
                     payload = "fluent"
                 );
                 let _guard = span.enter();
-                construct_block_cache_inner(&mut rng, &pyld, &block_chunks)?
+                construct_block_cache_inner(&mut rng, &pyld, &block_chunks, total_bytes.get())?
             }
             crate::Config::SplunkHec { encoding } => {
                 let span = span!(
@@ -239,6 +249,7 @@ impl Cache {
                     &mut rng,
                     &crate::SplunkHec::new(*encoding),
                     &block_chunks,
+                    total_bytes.get(),
                 )?
             }
             crate::Config::ApacheCommon => {
@@ -250,7 +261,7 @@ impl Cache {
                     payload = "apache-common"
                 );
                 let _guard = span.enter();
-                construct_block_cache_inner(&mut rng, &pyld, &block_chunks)?
+                construct_block_cache_inner(&mut rng, &pyld, &block_chunks, total_bytes.get())?
             }
             crate::Config::Ascii => {
                 let pyld = crate::Ascii::new(&mut rng);
@@ -261,7 +272,7 @@ impl Cache {
                     payload = "ascii"
                 );
                 let _guard = span.enter();
-                construct_block_cache_inner(&mut rng, &pyld, &block_chunks)?
+                construct_block_cache_inner(&mut rng, &pyld, &block_chunks, total_bytes.get())?
             }
             crate::Config::DatadogLog => {
                 let serializer = crate::DatadogLog::new(&mut rng);
@@ -272,7 +283,12 @@ impl Cache {
                     payload = "datadog-log"
                 );
                 let _guard = span.enter();
-                construct_block_cache_inner(&mut rng, &serializer, &block_chunks)?
+                construct_block_cache_inner(
+                    &mut rng,
+                    &serializer,
+                    &block_chunks,
+                    total_bytes.get(),
+                )?
             }
             crate::Config::Json => {
                 let span = span!(
@@ -282,7 +298,12 @@ impl Cache {
                     payload = "json"
                 );
                 let _guard = span.enter();
-                construct_block_cache_inner(&mut rng, &crate::Json, &block_chunks)?
+                construct_block_cache_inner(
+                    &mut rng,
+                    &crate::Json,
+                    &block_chunks,
+                    total_bytes.get(),
+                )?
             }
             crate::Config::Static { ref static_path } => {
                 let span = span!(
@@ -296,6 +317,7 @@ impl Cache {
                     &mut rng,
                     &crate::Static::new(static_path)?,
                     &block_chunks,
+                    total_bytes.get(),
                 )?
             }
             crate::Config::OpentelemetryTraces => {
@@ -307,7 +329,7 @@ impl Cache {
                     payload = "otel-traces"
                 );
                 let _guard = span.enter();
-                construct_block_cache_inner(rng, &pyld, &block_chunks)?
+                construct_block_cache_inner(rng, &pyld, &block_chunks, total_bytes.get())?
             }
             crate::Config::OpentelemetryLogs => {
                 let pyld = crate::OpentelemetryLogs::new(&mut rng);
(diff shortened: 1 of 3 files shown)


lading_payload/src/block.rs
```diff
@@ -585,38 +607,55 @@ fn chunk_bytes<const N: usize>(
...
+            continue;
+        }
+
+        match construct_block(&mut rng, serializer, block_size) {
+            Ok(Some(block)) => {
+                bytes_remaining = bytes_remaining.saturating_sub(block.total_bytes.get());
+                block_cache.push(block);
+            }
+            Ok(None) => {
+                // Update the dynamic "floor" if the block size proves too small
```
scottopell, line 642: I don't understand the None vs Err split here. `construct_block` returns `None` if the request was valid, but a block could not be constructed.
This seems like it should trigger the `adjust_block_size` codepath, not record this size as the new floor.
The floor is initialized to 0, so searching for the lower bound this way doesn't make sense.

lading_payload/src/block.rs
```diff
@@ -585,38 +607,55 @@ fn chunk_bytes<const N: usize>(
...
+                min_viable_size = block_size;
+            }
+            Err(_) => {
```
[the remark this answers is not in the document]
blt: Hmm I think you're right. I am confused about what signals are in play here.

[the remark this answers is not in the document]
blt: Attempted to clarify in [c1273a4](https://github.com/DataDog/lading/pull/886/commits/c1273a43bda9dabce2a3f4ed9862723191765977). I think, in short, we should disallow the serializers from returning an empty block and remove the optional success return: block or error.

lading_payload/src/block.rs
```diff
@@ -585,38 +607,76 @@ fn chunk_bytes<const N: usize>(
...
+        if block_size <= min_viable_size {
+            continue;
+        }
+
+        // `construct_block` returns Ok(Some(..)) when the block is
+        // constructed. Easy. `Ok(None)` is returned if a serializer is unable
+        // to fill a block, but only some serializers behave this
+        // way. Technically it's _not_ an error condition and serializers have
+        // historically been allowed to construct empty blocks, although that's
+        // of little use these days. Vestigial, I guess, and we try to avoid
```
scottopell, line 651: I don't view this as historical or vestigial, rather this is an intrinsic aspect of payload generation. Payload generators are user-configurable, and therefore the constraints that need to be met when constructing a block are also user-configurable. 
In my view, since this is driven by user configuration, there is _always_ a chance that misconfiguration leads to unsatisfiable conditions and an empty block would be "produced".

lading_payload/src/block.rs
```diff
@@ -463,65 +485,73 @@ fn chunk_bytes<const N: usize>(
...
+                bytes_remaining = bytes_remaining.saturating_sub(block.total_bytes.get());
                 block_cache.push(block);
-                true
-            } else {
-                construct_block_failures += 1;
-                false
             }
-        });
+            Err(SpinError::EmptyBlock) => {
+                min_viable_size = block_size;
```
scottopell, line 525: this is confusing to me, but I think its mostly bc the name.
Generating an empty block means "This block size cannot be satisfied". So how is that block size a "minimal viable size"?
But the way `minimum_viable_size` is used, the way to think about this is "Viable block sizes will all be above this size"

scottopell, line 525: Due to the random nature of the generators, the same request of N bytes may fail once and then succeed the next time. I don't know how to work around that fact in the algorithm, but I think we should at least have a check here to only grow the `min_viable_size`.

```suggestion
                min_viable_size = std::min(block_size, min_viable_size);
```

lading_payload/src/block.rs
```diff
@@ -463,65 +485,73 @@ fn chunk_bytes<const N: usize>(
...
+    // until runtime. We take the user-provided blocks and the total byte
+    // objective and iterate over these, dynamically constructing new block
+    // sizes +/- 20% if the block construction fails because of insufficient
+    // space to serialize.
+    while bytes_remaining > 0 {
+        let block_size = rng.gen_range(min_viable_size..=max_viable_size);
+
+        match construct_block(&mut rng, serializer, block_size) {
+            Ok(block) => {
+                bytes_remaining = bytes_remaining.saturating_sub(block.total_bytes.get());
```
scottopell, line 521: shouldn't there be an update to min/max size here? If we are able to generate a block for this `block_size`, that signals that this `block_size` should be in the viable range.

lading_payload/src/block.rs
```diff
@@ -463,65 +485,73 @@ fn chunk_bytes<const N: usize>(
...
+        .expect("empty block chunks, catastrophic bug");
+
+    // Build out the blocks.
+    //
+    // Our strategy here is to keep track of the minimal viable size of a block
+    // -- `min_viable_size` -- as the "floor" for block sizes. Because the
+    // serialization format varies we can't know what the floor actually is
+    // until runtime. We take the user-provided blocks and the total byte
+    // objective and iterate over these, dynamically constructing new block
+    // sizes +/- 20% if the block construction fails because of insufficient
```
scottopell, line 514: stale comment? I don't see the code to do +/- 20% anymore.

lading_payload/src/block.rs
scottopell: And/or adjust randomly per the +-20% idea

lading_payload/src/block.rs
blt: Yep, stale. Will fix.

lading_payload/src/block.rs
blt: I don't believe so. The block is taken from the range [`min_viable_size`, `max_viable_size`] so by definition `block_size` is not below the minimum viable size and is already in the viable range.

lading_payload/src/block.rs
blt: Ah yeah that is an interesting point. Because the `block_size` is  [min_viable_size, max_viable_size] we're guaranteed that it's always >= `min_viable_size`, so I think a `min` operation here is not called for. The minimum viable size will never fall below its current level. There will be some iterations where it does not change but I think that's acceptable.

blt: No, it's a good point. It should be at least one greater than the failed `block_size`. Likewise I don't think the range should be inclusive of the max, else we risk min == max which implies a panic.

blt: Addressed in [ecf5db3](https://github.com/DataDog/lading/pull/886/commits/ecf5db3d3d8e0c600b6208ea900d285aa16a79ae)

blt: Reintroduced random fluctuation in [ecf5db3](https://github.com/DataDog/lading/pull/886/commits/ecf5db3d3d8e0c600b6208ea900d285aa16a79ae). We achieve a tighter lower bound on the minimum block size -- 62B in one run where 1.1Kb had been the norm.

lading_payload/src/block.rs
```diff
@@ -463,65 +485,87 @@ fn chunk_bytes<const N: usize>(
...
+            Err(SpinError::EmptyBlock) => {
+                // It might be that `block_size` could not be constructed
+                // because the size is too small or we just caught a bad
+                // break. We do know that there's some true minimum viable size
+                // out there for each serialization format and user
+                // configuration, but we can only guess at it. To avoid racing
+                // _too_ far off the minimum viable size we scale the block size
+                // by -75% -- an arbitrary figure -- and set that as the new
+                // minimum block size.
+                min_block_size = (f64::from(block_size) * 0.25) as u32;
```
scottopell, line 536: I think I'm just being dense, but I don't understand this.
The algorithm currently:
- Generates a random block size between min (starts at 0) and max (starts at N).
- If attempt to serialize into this block size SUCCEEDS
    - add the block and we're done.
- If attempt to serialize into this block size FAILS
    - we take 1/4 of the attempted size and set that as the new min.

This last bit of taking 1/4 of the failed size is the bit that doesn't click for me.

If this size was too small (ie, it failed), why does that lead to this size being scaled down and set as the new min? Don't we already know that this value is too small? So why do we make it even smaller?

I'm thinking of the lower edge case here where min starts at 0 and we then generate a block size of...0. then this resets the min to 0 again. and loops again.

Maybe this just doesn't matter? We will just loop through and keep resetting the min to tiny values, but then eventually a bigger value comes along that fails, and we set the min to 25% of that new, bigger failing value?

blt: > This last bit of taking 1/4 of the failed size is the bit that doesn't click for me.

The thinking is, we could do `block_size + 1` but it's possible that another attempt at `block_size` would succeed -- an "unlucky" failure -- or that `block_size` can never succeed -- a "true" failure. Consider if we're generating JSON, the true minimum size is two bytes but if you add a key/value pair to an object that takes its own space. 

> If this size was too small (ie, it failed), why does that lead to this size being scaled down and set as the new min? Don't we already know that this value is too small? So why do we make it even smaller?

The insight here is we don't know truly if the size is too small, just that it didn't succeed once. We're attempting to approximate the _actual_ true minimum _and_ avoid always generating in the range `[0, max_block_size]` because while that full range does eventually terminate it takes a long, long time if the `max_block_size` is large. So, we guess. If `block_size` didn't pass we _can_ set this as the minimum block size and for dogstatsd we eventually settle on a minimum block of about 1Kb. If we scale it down as in this commit we find a minimum block of 62B, but not scaling does work. 

> I'm thinking of the lower edge case here where min starts at 0 and we then generate a block size of...0. then this resets the min to 0 again. and loops again.

That does happen, yeah, with some low probability. We make an even pull across the selected range but, you're right, there are some duplicate loops we make.

> Maybe this just doesn't matter? We will just loop through and keep resetting the min to tiny values, but then eventually a bigger value comes along that fails, and we set the min to 25% of that new, bigger failing value?

I don't think it matters, not in practice. A loop like this is weird. We _could_ add extra logic on each iteration to avoid extra iterations but that per-iteration cost outweighs the waste work, so it's better just to be naive. The only time it matters is when the serializer takes ages to produce a new instance but I think to this point we've stuck to the notion that serializers have to be quick in producing a new instance. 

FWIW generating dogstatsd in the full range `[0, 8Kb]` for 500Mb takes ~2 minutes and this current algorithm does the same ~4 seconds.

scottopell (approved):

PR #889, opened 2024-05-03
Remove explicit blocks

blt:
### What does this PR do?

This PR follows up from #886 and removes the block configuration entirely, preferring instead to allow the users to configure a maximum block size. This is intended to simplfy the lading implementation and drive it closer to how users make use of lading in practice.


goxberry (approved):
Neat.

PR #891, opened 2024-05-10
Dogstatsd: generate values that include the max of inclusive config ranges

GeorgeHahn:
### What does this PR do?

Bugfix; dogstatsd config ranges generated values using an exclusive range rather than inclusive range.


goxberry (approved):

goxberry:
Nice catch!

PR #913, opened 2024-07-11
Observer errors are now fatal

scottopell:
### What does this PR do?
Adds error handling by exiting lading when the `Observer` hits an error condition

### DIFF
--- a/lading/src/bin/lading.rs
+++ b/lading/src/bin/lading.rs
@@ -421,6 +421,7 @@ async fn inner_main(
     }
 
     let mut tsrv_joinset = tokio::task::JoinSet::new();
+    let mut osrv_joinset = tokio::task::JoinSet::new();
     //
     // OBSERVER
     //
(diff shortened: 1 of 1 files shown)


lading/src/bin/lading.rs
```diff
@@ -428,7 +428,12 @@ async fn inner_main(
     if let Some(target) = config.target {
         let obs_rcv = tgt_snd.subscribe();
         let observer_server = observer::Server::new(config.observer, shutdown.clone())?;
-        let _osrv = tokio::spawn(observer_server.run(obs_rcv));
+        let _osrv = tokio::spawn(async {
+            match observer_server.run(obs_rcv).await {
+                Ok(()) => debug!("observer shut down successfully"),
+                Err(err) => warn!("observer failed with {:?}", err),
```
GeorgeHahn, line 434: This seems like a critical error. What if we made lading exit if this is hit?

scottopell: Good suggestion, I'll push a change that joins this in the same `select` as the rest of the servers

blt (approved):

PR #914, opened 2024-07-12
Uses perf_event to grab the instruction and cycle count for target process

scottopell:
### What does this PR do?

Adds 3 additional telemetry gauges for the target software:
- instruction count
- cycle count
- Cycles Per Instruction

### Motivation

https://research.google.com/pubs/archive/40737.pdf


blt:
I like this idea. If we have an interest in it, glad to see this PR re-open.

PR #916, opened 2024-07-15
Update to rust 1.79.0

scottopell:
### What does this PR do?

Updates the rust version used in this project. 
Also updates the debian base-image version for the docker base.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.75.0"
+channel = "1.79.0"
 profile = "minimal"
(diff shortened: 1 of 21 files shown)


blt (approved):

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.75.0"
+channel = "1.79.0"
```
GeorgeHahn, line 2: Could you also update the rust version in the dockerfile and add a comment that both should be updated in lockstep?

GeorgeHahn (approved):

PR #918, opened 2024-07-16
Observe cgroup information on Linux

blt:
### What does this PR do?

This commit reads the cgroup memory controller and pulls limit et al information for each observed pid. I've copied telemetry names directly from Linux without consideration; I am open to a prefix for this new data or a different scheme. I intend to follow on with pulls from the CPU controller as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- Retrieve memory information from cgroup controller for every pid observed on Linux.
 
 ## [0.22.0-rc1]
 ### Fixed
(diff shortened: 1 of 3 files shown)


goxberry (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- Retrieve memory, CPU information from cgroup controller for every pid observed on Linux.
```
scottopell, line 9: shouldn't the cgroups collection only work off of the parent pid? AFAIK we don't have nested cgroups, so the child processes' cgroup data should all be identical to the parent

blt: In our private runner rigs this is true today, yes, but it's not strictly speaking a requirement for lading. Also, even in our private enviroment a target might spawn its own sub-processes and put them into groups.

PR #941, opened 2024-08-16
Enable tagging of target metrics from prometheus and expvar endpoints by sub-agent name 

cmetz100:
### What does this PR do?

This PR adds another configuration option to prometheus and expvar target metrics enabling a user to tag all the metrics coming from that configuration with optional additional labels

```
target_metrics:
  - prometheus: #process agent telemetry 
      uri: "http://127.0.0.1:6062/telemetry"
      tags:
        sub_agent: "process"
  - prometheus: #core agent telemetry
      uri: "http://127.0.0.1:5000/telemetry"
      tags:
        sub_agent: "core"
```
### Motivation

Currently in lading we could configure several prometheus endpoints to listen on to collect target metrics. In the example case of the agent being the target we could configure separate prometheus target metrics to get the internal telemetry of different sub agents. This issue is that when we do this there are some duplicate metrics across the sub agents. For example both the core agent and process agent expose workloadmeta telemetry. We want a way to differentiate not just the duplicate metrics but all of them, so that we can investigate the performance of each sub-agent on its own. This extends to any target. It would be useful to have additional labels configured to differentiate between target metric sources. The use of this applies beyond the agent target. The ability to tag metrics from specific target metrics endpoints would allow more flexibility and observability for any target. 

### Related issues

[SMPTNG-422](https://datadoghq.atlassian.net/browse/SMPTNG-422)

### Additional Notes

Open Questions:
-Should I include two labels in the example in the changelog? We only know of use cases at the moment using one additional tag but it is not really clear from the code that you can configure multiple. I also wanted to emphasize that the user can create the tag name by setting whatever key value they want, in addition to setting the tag value. 

[SMPTNG-422]: https://datadoghq.atlassian.net/browse/SMPTNG-422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,11 +20,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fixes bugs in `smaps` parsing code that can result in under-counting RSS in
   the smaps view of the data.
-### Added
-- Retrieve memory, CPU information from cgroup controller for every pid observed on Linux.
-
-## [0.22.0-rc1]
-### Fixed
 - Target observer was not exposed through CLI.
 ### Changed
 - Now built using rust 1.79.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - prometheus: #core agent telemetry
         uri: "http://127.0.0.1:5000/telemetry"
         tags:
-          sub-agent: "core-agent"
-          any-label: "any-string-value"
+          sub_agent: "core"
+          any_label: "any-string-value"
   ```
 
 ## [0.22.0]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,26 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- Added ability to create tags for both expvar and prometheus target metrics specific to a single target_metrics configuration (example below shows prometheus metrics collected from the core agent and two additional tags created)
+  ```yaml
+  target_metrics:      
+    - prometheus: #core agent telemetry
+        uri: "http://127.0.0.1:5000/telemetry"
+        tags:
+          sub-agent: "core-agent"
+          any-label: "any-string-value"
+  ```
 
 ## [0.22.0]
 ### Fixed
 - Fixes bugs in `smaps` parsing code that can result in under-counting RSS in
   the smaps view of the data.
+### Added
+- Retrieve memory, CPU information from cgroup controller for every pid observed on Linux.
+
+## [0.22.0-rc1]
+### Fixed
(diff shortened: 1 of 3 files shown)


lading/src/target_metrics/prometheus.rs
```diff
@@ -316,6 +331,51 @@ mod tests {
...
+        dr.install().expect("failed to install recorder");
+
+        experiment_started.signal();
+
+        p.scrape_metrics().await;
+
+        snapshotter.snapshot().into_hashmap()
+    }
+
+    async fn run_scrape_and_parse_metrics_with_sub_agent_label(
```
scottopell, line 351: it appears that this is very similar to the above `run_scrape_and_parse_metrics`, is there a good way to combine these to avoid duplicating this code?

scottopell (changes requested):
I like the overall approach, lets change the `sub_agent` terminology to be something more generic. `lading` is intended to be "target agnostic", so the target could be the Datadog Agent, but it could also be Vector or any other arbitrary program.

It also appears that there may be missing files from this branch, I would guess there are changes incoming in `target_metrics/expvar.rs` as well as `lading.rs`

cmetz100:
> I like the overall approach, lets change the `sub_agent` terminology to be something more generic. `lading` is intended to be "target agnostic", so the target could be the Datadog Agent, but it could also be Vector or any other arbitrary program.
> 
> It also appears that there may be missing files from this branch, I would guess there are changes incoming in `target_metrics/expvar.rs` as well as `lading.rs`

I was thinking about expanding the functionality once i got this working to enable the addition of a arbitrary amount of key value labels so that we could configure appropriate label names for whatever target. And yeah working on the expvar changes right now.

lading/src/target_metrics/prometheus.rs
cmetz100: I could refactor the tests to have the prometheus configuration passed in from the test function rather than being defined in the scraping helpers yeah. Going to do this after I refactor to generalize labels to any target

scottopell (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+## [0.23.0-rc0]
+### Added
+- Added ability to create tags for both expvar and prometheus target metrics specific to a single target_metrics configuration (example below shows prometheus metrics collected from the core agent and two additional tags created)
+  ```yaml
+  target_metrics:      
+    - prometheus: #core agent telemetry
+        uri: "http://127.0.0.1:5000/telemetry"
+        tags:
+          sub-agent: "core-agent"
```
scottopell, line 17:
```suggestion
          sub_agent: "core"
```

small suggestion here to prefer `_` instead of `-` for consistence with our other metrics (eg, `job_id`) and to not duplicate the word "agent" in the value

cmetz100: makes sense to me

cmetz100: these changes really apply for the changes I will make in [this pr](https://github.com/DataDog/single-machine-performance/pull/2069) but I can update the examples here accordingly

PR #942, opened 2024-08-16
Improve smaps parsing support

scottopell:
### What does this PR do?

This fixes 2 bugs in the smaps parsing logic and makes 1 simplification.
1. Simpler method of reading smaps to a string, the size-hinting should be done by stdlib
2. A smaps region (VMA) can have varying hex address lengths, this was not properly accounted for
3. The logic to split an smaps file into individual memory regions (VMAs) needs to only match address-range looking things that appear at the beginning of the line.

### Motivation

Inaccurate `/proc/[pid]/smaps` data.

### DIFF
--- a/lading/src/observer/memory.rs
+++ b/lading/src/observer/memory.rs
@@ -368,7 +368,7 @@ impl Regions {
         // regions are separated by a line like this:
         // 7fffa9f39000-7fffa9f3b000 r-xp 00000000 00:00 0                          [vdso]
         let region_start_regex =
-            Regex::new("[[:xdigit:]]+-[[:xdigit:]]+[[:space:]]").expect("Regex to be valid");
+            Regex::new("(?m)^[[:xdigit:]]+-[[:xdigit:]]+").expect("Regex to be valid");
         let mut start_indices = region_start_regex.find_iter(contents).map(|m| m.start());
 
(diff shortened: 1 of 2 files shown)


lading/src/observer/memory.rs
```diff
@@ -56,7 +55,7 @@ impl Rollup {
     pub(crate) fn from_file(path: &str) -> Result<Self, Error> {
         let mut file: std::fs::File = std::fs::OpenOptions::new().read(true).open(path)?;
 
-        let mut contents = String::with_capacity(SMAP_SIZE_HINT);
+        let mut contents = String::new();
```
blt, line 58: I'm curious about the thinking here, why we'd want to drop the size hint.

blt (approved):

lading/src/observer/memory.rs
scottopell: from what I was reading, `read_to_end` and `read_to_string` will do a `try_reserve` with a size-hint of the actual file size. I figured that's sufficient.

scottopell: https://doc.rust-lang.org/stable/src/std/fs.rs.html#822-833

blt: Oh yeah, good call.

PR #965, opened 2024-08-30
Split lading::signals::Phase into lading_signals

blt:
### What does this PR do?

We know that in practice we have a problem with our Phase mechanism. It had become overburdened with the sense of multiple transitions being possible and it was impossible to test. This commit moves the Phase mechanism to its own crate and our intention is to loom test it. To achieve this the internal complexity has been significantly reduced, changing the API slightly to rely on ownership, drops to ensure -- hopefully -- that the necessary semantics of a single signal transition is maintained.

Loom tests will help us determine if this is true.

### Motivation

REF SMPTNG-455


lading_signal/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+            Err(broadcast::error::RecvError::Lagged(_)) => {
+                panic!("Catastrophic programming error: lagged behind");
+            }
+        }
+    }
+
+    /// Check if a shutdown notice has been sent without blocking.
+    ///
+    /// If the signal has not been received returns Ok(false). If it has been
+    /// received Ok(true). All calls after will return `TryRecvError::SignalReceived`.
```
scottopell, line 169: Is this useful behavior? Naively I would want all subsequent calls to continue to return `Ok(true)`

scottopell: This is also different than the async version of `recv`

lading_signal/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+    /// Send the signal through to any `Watcher` instances.
+    ///
+    /// Function WILL block until all peers have ack'ed the signal.
+    #[tracing::instrument(skip(self))]
+    pub async fn signal_and_wait(self) {
+        drop(self.sender);
+
+        // Wait for all peers to drop off.
+        while self.peers.load(Ordering::SeqCst) > 0 {
+            self.notify.notified().await;
```
scottopell, line 83: Note: This is the other side of `decrease_peer_count`. Usage of `Notify` is effectively a semaphore with peers Adding a permit when they ack the notification.

lading_signal/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+
+#[derive(Debug)]
+/// Mechanism to watch for phase changes, typically used to control shutdown.
+pub struct Watcher {
+    /// Used to track if the signal has been received without synchronization.
+    signal_received: bool,
+    /// Record whether the peer count of this Watcher has been decreased.
+    peer_count_decreased: bool,
+    /// The total number of peers subscribed to the `Broadcaster`.
+    peers: Arc<AtomicU32>,
```
scottopell, line 112: I don't follow why the Watcher needs to know what peers exist.
Isn't each instance of a Watcher a single peer to the broadcaster? The broadcaster needs to know about all watchers, but why do watchers need to know about other watchers?

lading_signal/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+    /// Register with the `Broadcaster`, returning a new instance of `Watcher`.
+    #[tracing::instrument(skip(self))]
+    pub fn register(&self) -> Result<Self, RegisterError> {
+        if self.signal_received {
+            // If the shutdown signal has already been received, return with
+            // error.
+            return Err(RegisterError::SignalReceived);
+        }
+
+        self.peers.fetch_add(1, Ordering::SeqCst);
```
scottopell, line 199: Ah, is this why the watcher needs to know about peers? Because the interface allows Watchers to create more watchers?

If that's the case, can we get away with a simpler approach that only allows Watchers to be created from Broadcasters?

lading_signal/src/lib.rs
blt: Ah yeah that's a good comment to add in here.

lading_signal/src/lib.rs
blt: Hmm, we could make the behavior the same if we wanted, pending the loom tests passing of course. The idea I had in mind -- and it was late -- was that this would force the peer to bale out, but you are correct. It's a different behavior.

lading_signal/src/lib.rs
blt: They don't, but each individual `Watcher` must be able to inform the `Broadcaster` of its existence, lack of. Keeping a counter is a convenient way to achieve that without allocation or similar.

lading_signal/src/lib.rs
blt: We can't. Or, not conveniently. The idea with the Phase mechanism is that you can keep passing `Phase` downward and clone as you need. Splitting the broadcast/watch idea apart we preserve that "clone" behavior on the `Watcher` but the `Broadcaster` only exists at one point. 

I suppose if we really wanted to we could put the broadcaster in an arc and pass that around, but you'd be either forever registering yourself a watcher off it. This also echos how broadcast channels tend to work, where the subscription is hung off the receiver.

lading_signal/src/lib.rs
blt: Ah looking at this with fresh eyes this behavior is important as it keeps the CaptureManager from spinning forever due to a programming mistake. I think actually following up this PR with one that switches the semantics of `recv` to `try_recv` would be a solid idea: the fast return of `recv` means a `select!` could potentially just hit on that infinitely. I'll stick a yield in it.

scottopell (approved):

PR #981, opened 2024-09-06
Update rust toolchain to 1.81.0

blt:
### What does this PR do?

This commit updates the rust toolchain in lading project to latest stable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Update the rust version in-sync with the version in rust-toolchain.toml
-FROM docker.io/rust:1.79.0-bullseye as builder
+FROM docker.io/rust:1.81.0-bullseye as builder
 
 RUN apt-get update && apt-get install -y \
     protobuf-compiler \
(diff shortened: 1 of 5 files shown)


GeorgeHahn (changes requested):
The docker images take some time to be published. I retried the build that was previously failing to find the 1.81.0 image and it made it past that step: https://github.com/DataDog/lading/actions/runs/10730774766/job/29793904675. I think this would be worth updating before merge to keep the compiler version consistent.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Update the rust version in-sync with the version in rust-toolchain.toml
-FROM docker.io/rust:1.79.0-bullseye as builder
+FROM docker.io/rust:1.79.0-bullseye AS builder
```
scottopell, line 2:
```suggestion
FROM docker.io/rust:1.81.0-bullseye AS builder
```

GeorgeHahn (approved):

PR #1007, opened 2024-09-24
Capture files will now exclude metrics that have not been emitted recently

scottopell:
### What does this PR do?

Only consider metrics emitted within the last 2 seconds when emitting to the capture file

### Motivation

Stale counters are a particularly large problem when considering short-lived processes and the observer metrics.

### Related issues

This is basically the exact same code from #748 , which was later reverted in #762. At the time, a deadlock was occurring which prevented forward progress.

That is the main risk I'll be evaluating while trying to land this change.

### Additional Notes

Currently this relies on a custom branch of `metrics-rs` https://github.com/metrics-rs/metrics/pull/522

This is needed due to the `RWMutex` interactions with `Recency`, so another option for this is to rewrite the capture exporter entirely to avoid needing `Recency` at all.

```
George Hahn
  Yesterday at 1:21 PM
Could we get onto a better path if we moved away from the prometheus exporter? We copied and customized it to write a file, and we should have flexibility with how that's wired up.
tobz
100%, yes
tobz
my understanding of your metrics load is that it's not very high -- like 100s to 1000s of emissions per second? -- which would be more than fast enough even if you just did a simple Mutex<HashMap<...>> or whatever
:nod2:

tobz
  and then every reporting period consumes that map, and voila, now you're "expiring" old metrics
George Hahn
  Yep, that's right in line with our typical load. We handle writing on a separate thread too, bet we could swap maps and block for almost no time.
tobz
  yeah, swap maps, whatever
Scott Opell
  I don’t follow the idea here, are we talking about a custom Registry? Or just calling retain_* to delete all metrics after a given reporting peried?
George Hahn
  I should note I'm not up to speed on metrics internals. My thought is that we could fully break from the patterns used in the capture manager today and build a recorder from the ground up. Looking at something like the [tcp exporter](https://github.com/metrics-rs/metrics/blob/main/metrics-exporter-tcp/src/lib.rs), I think that could mean running without a registry at all.
tobz
  yeah, exactly
tobz
  it could be anything from:
every every handle just wraps an Arc<Mutex<HashMap<...>>> and does a get/insert to emit a metric
every handle wraps a Arc<SomeMpscChannel<...>> and sends a message to emit a metric
tobz
  or something else!
```

REF [SMPTNG-85](https://datadoghq.atlassian.net/browse/SMPTNG-85)

[SMPTNG-85]: https://datadoghq.atlassian.net/browse/SMPTNG-85?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1434,6 +1434,7 @@ dependencies = [
  "procfs",
  "proptest",
  "proptest-derive",
+ "quanta",
  "rand",
  "regex",
  "reqwest",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -704,7 +704,7 @@ dependencies = [
  "once_cell",
  "serde_json",
  "shared",
- "sketches-ddsketch 0.3.0",
+ "sketches-ddsketch",
  "tokio",
  "tokio-stream",
  "tonic 0.9.2",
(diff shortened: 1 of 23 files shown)


scottopell:
This approach isn't what we want to go with, for more details see https://datadoghq.atlassian.net/browse/SMPTNG-85

PR #1026, opened 2024-10-04
Default response latency for blackhole http requests 0 -> 30ms

scottopell:
### What does this PR do?

Changes the default response latency for http blackhole requests.

### Motivation
 
0ms is not a realistic or useful latency value. Target software will basically never encounter 0ms responses.

### Additional Notes

An argument could be made that this is a breaking change, it has a new default for all experiments using the http blackhole (ie, all of them)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Linux observer is more resilient to scenarios where lading lacks ptrace permission.
+- Default response latency for HTTP blackhole is now 30ms (previously 0ms)
 ### Removed
 - lading_capture no longer exports a protobuf version of the capture.
 ### Added
--- a/lading/src/blackhole/http.rs
+++ b/lading/src/blackhole/http.rs
@@ -63,7 +63,7 @@ fn default_body_variant() -> BodyVariant {
 }
 
 fn default_response_delay_millis() -> u64 {
-    0
+    30
 }
 
 fn default_status_code() -> u16 {


remeh (approved):

blt (changes requested):
I'm concerned about putting modeling assumptions into lading.

scottopell:
My take on this is that we're already modeling this in lading, we're just modeling it with an unrealistic value.

I'm open to making this a two-step upgrade, next lading version can expose this value as configurable and any experiments concerned with the latency can set it to `0` to preserve existing behavior. Then next-next lading version can set this as the default.

blt:
Sure, but it's intentional. It's why the blackhole is built to minimize latency to response to begin with. I feel strongly that we shouldn't model realism in lading by default but strive to continue with the design goal of lading in generation and sink providing zero overhead unless specifically requested. 

I could get behind making intake response delay not optional but making a non-zero default is overfitting to a specific target.

goxberry:
re: zero latency, I'd be in favor of keeping it as-is. There's a strong tradition in scientific and mathematical modeling of starting with a model that reflects "ideal" -- possibly unrealistically ideal -- conditions, and then adding correction terms where needed. If Agent needs a more realistic model, changing latency via a setting seems reasonable to me.

I view this scenario as different from the discussions re: discarding data in our statistical analyses because I think I view the statistical analysis discussion there as a discussion of whether a metric's value is essentially impossible, not merely "unrealistically ideal".

scottopell:
yeah I can get behind this, keep lading as the configurable "base-layer" and then any modeling decisions are applied at a higher abstraction layer. In the current world, its just a lot of hand-edited yaml. 

There is an opening here for a higher-level abstraction where a user could configure "http blackhole over internet" that models latency/jitter that is realistic for traffic rount-tripping through the public internet.

blt:
> There is an opening here for a higher-level abstraction where a user could configure "http blackhole over internet" that models latency/jitter that is realistic for traffic round-tripping through the public internet.

That'd be slick.

PR #1031, opened 2024-10-07
Adds ability for payload tool to emit the raw contents of blocks

scottopell:
### What does this PR do?

Adds another option to payloadtool to debug block generation

### Motivation

Useful to see the actual output of some lading payloads, ie, dogstatsd.


blt:
@scottopell I think this could still be interesting if you're of a mind to pursue it. Else, I'll close.

blt (approved):

PR #1032, opened 2024-10-07
When generating dsd metrics, always attempt to use each context once

scottopell:
### What does this PR do?

Changes the way that dogstatsd metrics generation works by attempting to always use all contexts

### Motivation

It is surprising and unintuitive to configure lading and _not_ see the expected traffic. This helps.


blt:
@scottopell keep or close do you think? Also CC @tobz .

blt:
I like this idea. If we have an interest in it, glad to see this PR re-open.

PR #1054, opened 2024-10-22
Send the HTTP Host header for the HTTP generator.

dustmop:
### What does this PR do?

Send the HTTP Host header for the HTTP generator and the Splunk generator.

### Motivation

The Host header is required for compliant HTTP traffic. Send it so that standard web servers will properly receive HTTP requests.

Note that there are two other places in the code that this `set_host(false)` method is used ([1](https://github.com/DataDog/lading/blob/4f7ec01c17fcf74abcf55cc9d9226a08c6155ff3/lading/src/generator/splunk_hec.rs#L251)) ([2](https://github.com/DataDog/lading/blob/4f7ec01c17fcf74abcf55cc9d9226a08c6155ff3/lading/src/generator/splunk_hec/acknowledgements.rs#L92)). I'm leaving them as is because I'm not familiar with Splunk and whether it needs this behavior.

UPDATE: Commit 9d3e627534d964 now removes the `set_host(false)` call from Splunk also, so Host headers are added by `hyper` everywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   them rather than incorrectly identifying the metric value.
 - Prometheus target metrics scraper will no longer panic if a metric has an
   invalid value (instead it will be logged)
-- HTTP traffic sent by the HTTP generator and Splunk generator now always include
+- HTTP traffic sent by the HTTP generator and Splunk generator now always includes
   a Host header to comply with HTTP 1.1 requirements
 
 ## [0.23.3]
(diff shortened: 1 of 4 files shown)


jszwedko (commented):
This is a minor change, but I think it still warrants a changelog entry. I defer to @DataDog/single-machine-performance though.

scottopell (changes requested):
Could you add a changelog entry please?

dustmop:
Changelog added.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   them rather than incorrectly identifying the metric value.
 - Prometheus target metrics scraper will no longer panic if a metric has an
   invalid value (instead it will be logged)
+- HTTP traffic sent by the HTTP generator and Splunk generator now always include
```
scottopell, line 13:
```suggestion
- HTTP traffic sent by the HTTP generator and Splunk generator now always includes
```

scottopell (approved):

PR #1076, opened 2024-10-29
Avoid some allocations, hashes that aren't needed

blt:
### What does this PR do?

This commit removes allocations in the FUSE component that don't
need to exist and adjust the interior hashmap to use FxHashMap,
avoiding a secure hash in favor of a fast one. Also, directories
now keep their children in a BTreeSet so that readdir is always
in the same order when called.

### DIFF
--- a/examples/lading-logrotatefs.yaml
+++ b/examples/lading-logrotatefs.yaml
@@ -8,7 +8,7 @@ generator:
         total_rotations: 4
         max_depth: 0
         variant: "ascii"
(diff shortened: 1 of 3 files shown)


goxberry (approved):

blt:
### Merge activity

* **Oct 31, 3:03 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/DataDog/lading/1076).

PR #1088, opened 2024-11-04
Attempt to speed up container builds

blt:
### What does this PR do?

The lading container build right now is roughly 1 hour. This is a very long loop to wait on. Introduce cargo-chef and more aggressive caching to hopefully cut this time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ WORKDIR /app
 RUN apt-get update && apt-get install -y \
     protobuf-compiler fuse3 libfuse3-dev \
     && rm -rf /var/lib/apt/lists/*
-RUN cargo install cargo-chef
+RUN cargo install cargo-chef --locked
 
 # Planning stage, artifact is 'recipe.json'
 FROM chef AS planner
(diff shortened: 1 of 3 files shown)


GeorgeHahn:
As a smoke test, could you push a small code change?

Dockerfile
```diff
@@ -1,18 +1,36 @@
...
+# Lading arm64 builds are very slow. We use cargo-chef and aggressive layer
+# caching to avoid some of the build delay.
 
+# Builder image, cargo-chef installed
+FROM docker.io/rust:1.81.0-bullseye AS chef
+WORKDIR /app
 RUN apt-get update && apt-get install -y \
     protobuf-compiler fuse3 libfuse3-dev \
     && rm -rf /var/lib/apt/lists/*
+RUN cargo install cargo-chef
```
GeorgeHahn, line 10: This should probably use the project lockfile, `--locked`

blt: Addressed in [759c7c8](https://github.com/DataDog/lading/pull/1088/commits/759c7c82d4945df7b4521c38ce17cd70ed0fe28e)

scottopell:
Does stripping the release binary help with compile time? I think its useful to preserve debug symbols as a general practice, unless there's a compelling reason.

blt:
> Does stripping the release binary help with compile time? I think its useful to preserve debug symbols as a general practice, unless there's a compelling reason.

Yeah it shaves ~4 minutes off. The lading binary has gotten fairly big. 

On the whole, I'm meh on this change I think. It complicates the build non-trivially and it doesn't really improve build speed. An ARM build machine is what we need.

PR #1089, opened 2024-11-04
Remove release workflow

blt:
### What does this PR do?

We consider lading's container to be the only supported release artifact at this time. If a use-case for binaries or the crates.io artifacts appear in the future we'll be glad to re-introduce them.


.github/workflows/release.yml
```diff
@@ -1,103 +0,0 @@
...
-  create-release:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-
-      - name: Create GitHub release
-        uses: taiki-e/create-gh-release-action@72d65cee1f8033ef0c8b5d79eaf0c45c7c578ce3 # v1.8.2
-        with:
-          changelog: CHANGELOG.md
-          token: ${{ secrets.GITHUB_TOKEN }}
```
GeorgeHahn, line 21: Thoughts on retaining this piece? I find the changelogs helpful at least.

.github/workflows/release.yml
```diff
@@ -1,103 +0,0 @@
...
-
-  upload-binaries:
```
GeorgeHahn, line 23: Could you also remove the readme note about binaries? https://github.com/DataDog/lading?tab=readme-ov-file#binary-releases

GeorgeHahn: Looking at this again, it'd be helpful to have a link to lading on GHCR: https://github.com/DataDog/lading/pkgs/container/lading.

.github/workflows/release.yml
blt: I'll put this back.

blt: Resolved in [bb68f5e](https://github.com/DataDog/lading/pull/1089/commits/bb68f5e08a442a68087aafd9379cc7a966eb7c50)

GeorgeHahn (approved):

.github/workflows/release.yml
GeorgeHahn: Ideal!

PR #1092, opened 2024-11-05
Add load profile to logrotate FS

blt:
### What does this PR do?

We have a need for some experiments to ramp their load up over time.
Current configuration is constant over the lifetime of the experiment,
this change introduces linear growth in addition to constant.

A new release version will be required as this is a breaking change to
logrotate_fs configuration.

REF SMPTNG-517


blt:
* **#1092** <a href="https://app.graphite.dev/github/pr/DataDog/lading/1092?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @blt and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

lading/src/generator/file_gen/logrotate_fs/model.rs
```diff
@@ -281,6 +272,20 @@ pub(crate) enum Node {
     },
 }
 
+/// Profile for load in this filesystem.
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum LoadProfile {
```
scottopell, line 277:
```suggestion
pub(crate) enum LoadProgression {
```

My thought is that the name should reflect that this defines how the load changes over time. "Profile" doesn't imply "over time", whereas progression does.

I also like `LoadGrowth`, and `Constant` would be `None` (no growth, just constant)

scottopell (approved):
Looks good to me, only one naming-related comment

lading/src/generator/file_gen/logrotate_fs/model.rs
blt: I chose profile because I thought it was sort of neutral in the sense that we could incorporate things that don't change over time or don't have a sensible growth curve, like something that's just random.

This also made me think the block configuration should be wrapped up in a common spot and not repeated in every generator.

blt:
### Merge activity

* **Nov 5, 6:31 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/DataDog/lading/1092).

PR #1099, opened 2024-11-06
Use native ARM builders for container builds

blt:
### What does this PR do?

This commit replaces #1088 and attempts to build a lading multi-platform image by use of organization provided ARM native builders. Unsure if this'll work but I figure it's worth a shot.

### DIFF
--- a/.github/actionlint.yml
+++ b/.github/actionlint.yml
@@ -1,0 +1,3 @@
+self-hosted-runner:
+  labels:
+    - arm-4core-linux-ubuntu24.04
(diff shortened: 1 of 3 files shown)


scottopell (approved):
nice! 
Do we need any changes to the metadata tagging to ensure that the manifest gets tagged with the version number during release?

blt:
> nice! Do we need any changes to the metadata tagging to ensure that the manifest gets tagged with the version number during release?

I don't think so but let me reconfirm and make a PR if that's broke.

PR #1105, opened 2024-11-11
Add support for transparent huge pages in linux memory observer

scottopell:
### What does this PR do?

Parses out an additional field for `smaps` and `smaps_rollup` to see how much RSS is being consumed by https://docs.kernel.org/admin-guide/mm/transhuge.html

### Motivation

This could potentially explain some of the variance observed when running identical Agent processes.


blt (approved):

scottopell:
`smaps_rollup` is broken here somehow, not exactly sure how.

scottopell:
superceded by #1120

PR #1113, opened 2024-11-12
Validate operations on FileHandles

blt:
### What does this PR do?

This commit allows State to track whether a file handle is valid for read etc
operations. Our GC mechanism keeps track of the open-handles but assumes the
correctness of callers, meaning if multiple calls to `release` are made at the
filesystem level a node may be GC'ed before its time.

The State now keeps a record of which FileHandles exist and whether they remain
valid, meaning the GC mechanism cannot be misled.

### Motivation

SMPTNG-531

### DIFF
--- a/lading/src/generator/file_gen/logrotate_fs/model.rs
+++ b/lading/src/generator/file_gen/logrotate_fs/model.rs
@@ -298,6 +298,7 @@ pub(crate) struct State {
     next_file_handle: u64,
     inode_scratch: Vec<Inode>,
     load_profile: LoadProfile,
+    valid_file_handles: FxHashMap<u64, Inode>, // Track valid FileHandle IDs -> Inode
 }
 
 impl std::fmt::Debug for State {
@@ -387,6 +388,7 @@ impl State {
             next_file_handle: 0,
             inode_scratch: Vec::with_capacity(concurrent_logs as usize),
             load_profile,
+            valid_file_handles: FxHashMap::default(),
         };
 
         if concurrent_logs == 0 {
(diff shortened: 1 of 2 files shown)


blt:
* **#1113** <a href="https://app.graphite.dev/github/pr/DataDog/lading/1113?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @blt and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

lading/src/generator/file_gen/logrotate_fs/model.rs
```diff
@@ -1240,6 +1255,51 @@ mod test {
...
+                                } else {
+                                    None
+                                }
+                            })
+                            .collect();
+
+                        assert!(
+                        !valid_handles.is_empty(),
+                        "Unlinked, read-only file with open handles should have valid file handles"
+                    );
```
GeorgeHahn, line 1292: Nit: indentation is off here

GeorgeHahn (approved):

lading/src/generator/file_gen/logrotate_fs/model.rs
blt: So it is and cargo-fmt didn't catch it.

GeorgeHahn: Macros :( There was a google summer of code project that I think will improve this in the next few releases.

PR #1125, opened 2024-12-04
Attempt generational storage of capture metrics

blt:
### What does this PR do?

This commit attempts the generational storage of metrics in the capture
recorder. I am patterning this work from metrics-prometheus-exporter and
I think I understand how this is wired up?

One thing of note gauges are now ints and not f64s although the interface
still requires f64 on the gauge! side. Unsure how that works out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - logrotate_fs is now behind a feature flag and not enabled in the default
   build. It remains enabled in the release artifact.
 - The build now includes http1 and http2 support. Actual usage and availability may vary.
-- Metrics storage is now generational, expiring unwritten metrics every 7 seconds.
+- Metrics storage is now generational, expiring unwritten metrics every 3 seconds.
 
 ## [0.24.0]
 ## Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - logrotate_fs is now behind a feature flag and not enabled in the default
   build. It remains enabled in the release artifact.
 - The build now includes http1 and http2 support. Actual usage and availability may vary.
+- Metrics storage is now generational, expiring unwritten metrics every 7 seconds.
 
 ## [0.24.0]
 ## Added
--- a/lading/src/captures.rs
+++ b/lading/src/captures.rs
@@ -88,7 +88,7 @@ impl CaptureManager {
             recency: Recency::new(
                 quanta::Clock::new(),
                 MetricKindMask::GAUGE | MetricKindMask::COUNTER,
-                Some(Duration::from_secs(7)),
+                Some(Duration::from_secs(3)),
             ),
         };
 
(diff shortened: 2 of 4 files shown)


blt:
* **#1125** <a href="https://app.graphite.dev/github/pr/DataDog/lading/1125?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/lading/1125?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - logrotate_fs is now behind a feature flag and not enabled in the default
   build. It remains enabled in the release artifact.
 - The build now includes http1 and http2 support. Actual usage and availability may vary.
+- Metrics storage is now generational, expiring unwritten metrics every 7 seconds.
```
scottopell, line 12:
```suggestion
- Metrics storage is now generational, expiring unread metrics every 7 seconds.
```

scottopell (commented):
We can assert that the registry will be emptied and fully consumed once per second, I'm thinking we could take advantage of that fact and avoid the use of "Recency" altogether

blt:
> We can assert that the registry will be emptied and fully consumed once per second, I'm thinking we could take advantage of that fact and avoid the use of "Recency" altogether

I think we do have to make use of recency. The registry isn't consumed when we trawl it in record_captures and if I read the documentation correctly the generational mechanism is what's used to expire metrics. If we don't use recency we're racing. 

Haven't read the metrics-rs code to confirm.

CHANGELOG.md
blt: I believe it's based on writes, at least looking at the docs.

lading/src/captures.rs
```diff
@@ -42,7 +46,8 @@ pub enum Error {
 }
 
 struct Inner {
-    registry: Registry<metrics::Key, AtomicStorage>,
+    registry: Registry<Key, GenerationalAtomicStorage>,
+    recency: Recency<Key>,
```
GeorgeHahn, line 50: Is timed expiry giving us a benefit here? [SMPTNG-85](https://datadoghq.atlassian.net/browse/SMPTNG-85) has notes on replacing this with a map that can be consumed every second.

Considerations:

Switching to a store that is consumed every second would change values to be diff values rather than the accumulated values we currently record. That would require a rouster change to handle along with a flag to let it know how to process incoming captures. The compat side is annoying but medium-term this would lead to simpler code here and in rouster.

There's an edge case here that will trip rouster up: if a context stops being reported and is later reported again, I'm pretty sure we'll get wrong values out of rouster's diff calculation because it is unaware of expiry.

blt: I think that would be a valuable thing to plan out in the future, but it's outside scope for the stack I'm working on, considering we'd be talking about a non-trivial amount of coordination.

GeorgeHahn: Re-reading this, I realize I misunderstood the rouster impact. I was thinking contexts were reset after expiry; that doesn't look to be the case. This looks good from the rouster side.

GeorgeHahn (approved):

blt:
### Merge activity

* **Dec 5, 2:57 AM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/DataDog/lading/1125).

PR #1178, opened 2024-12-19
Reformat procfs scraping to handle errors

cmetz100:
### What does this PR do?

Extracts all the per-pid operations into a helper function so that if we cant read something for a given pid it doesnt blow up the whole observer.

### Motivation

Processes exiting mid poll loop can cause numerous errors, all of which, if not handled properly, polute the collection of data from active processes.

### Related issues

First attempts at this in https://github.com/DataDog/lading/pull/1175 and https://github.com/DataDog/lading/pull/1142
After some discussion we thought it better to just catch all per-pid related errors rather than keeping up with new errors on individual functions related to collecting per process information. 

### Additional Notes

### DIFF
--- a/lading/src/observer/linux/procfs.rs
+++ b/lading/src/observer/linux/procfs.rs
@@ -118,158 +118,175 @@ impl Sampler {
                     }
                 }
             }
+            // Update the process_info map to only hold processes seen by the current poll call.
+            self.process_info.retain(|pid, _| pids.contains(pid));
 
             let pid = process.pid();
-
-            // `/proc/{pid}/status`
-            let status = match process.status() {
-                Ok(status) => status,
-                Err(e) => {
-                    warn!("Couldn't read status: {:?}", e);
-                    // The pid may have exited since we scanned it or we may not
-                    // have sufficient permission.
-                    continue;
-                }
-            };
-            if status.tgid != pid {
-                // This is a thread, not a process and we do not wish to scan it.
-                continue;
-            }
-
-            // If we haven't seen this process before, initialize its ProcessInfo.
-            match self.process_info.entry(pid) {
-                Entry::Occupied(_) => { /* Already initialized */ }
-                Entry::Vacant(entry) => {
-                    let exe = proc_exe(pid).await?;
-                    let comm = proc_comm(pid).await?;
-                    let cmdline = proc_cmdline(pid).await?;
-                    let pid_s = format!("{pid}");
-                    let stat_sampler = stat::Sampler::new();
-
-                    entry.insert(ProcessInfo {
-                        cmdline,
-                        exe,
-                        comm,
-                        pid_s,
-                        stat_sampler,
-                    });
-                }
-            }
-
-            // SAFETY: We've just inserted this pid into the map.
-            let pinfo = self
-                .process_info
-                .get_mut(&pid)
-                .expect("catastrophic programming error");
-
-            let labels: [(&'static str, String); 4] = [
-                ("pid", pinfo.pid_s.clone()),
-                ("exe", pinfo.exe.clone()),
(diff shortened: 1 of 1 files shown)


lading/src/observer/linux/procfs.rs
```diff
@@ -118,158 +118,182 @@ impl Sampler {
...
+        clippy::too_many_lines,
+        clippy::cast_sign_loss,
+        clippy::cast_possible_truncation,
+        clippy::cast_possible_wrap
+    )]
+    async fn handle_process(
+        &mut self,
+        process: Process,
+        aggr: &mut memory::smaps_rollup::Aggregator,
+    ) -> Result<(), Error> {
```
scottopell, line 148: Is there any codepath that returns an error? I'm not seeing any.

My thought here is that we need to explicitly handle errors when reading process information, ie, avoid the `?` operator. So if we make this process not have a return type, we could avoid any accidental error handling that uses `?`

cmetz100: Hm yeah I think I see what you are saying. I chose this implementation specifically to handle `?` without having to explicitly handle errors. Mostly just because the proc_exe(), proc_comm(), and proc_cmdline() all can error if the pid had exited and I thought it was cleaner to just blanket catch any of those failures rather then add messiness catching different failures that are all resulting from the same root cause (the pid exiting)

cmetz100: Very open to changing this tho to catch every error explicitly, just thought this may be easier to add other code later in handle_process without having to risk observer crashes

scottopell: > proc_exe(), proc_comm(), and proc_cmdline() all can error if the pid had exited and I thought it was cleaner to just blanket catch any of those failures rather then add messiness catching different failures that are all resulting from the same root cause (the pid exiting)

What makes these failures different from, eg `process.status()` failing?

In my understanding, any and all errors that could be caused by a process exiting during execution of `handle_process` should be explicitly caught and handled by ignoring and `return`/`return Ok(())`.

If there are errors that can occur that are _not_ caused by the process exiting early, then those would be a valid reason to bubble an error out of this function, but I don't think process-exit-errors should escape this function, otherwise what is the point of extracting into `handle_process`?

cmetz100: Ah yeah I see what you are saying. It is the same type of failure as `process.status()`. I will update the code to catch what I expect to be specific "pid-exit" errors. The only exception to that I see is the uptime poll since that is more broad than a specific process

lading/src/observer/linux/procfs.rs
```diff
@@ -118,158 +118,205 @@ impl Sampler {
...
-                }
-            };
-            if status.tgid != pid {
-                // This is a thread, not a process and we do not wish to scan it.
-                continue;
+            if let Err(e) = self.handle_process(process, &mut aggr).await {
+                warn!("Encountered uncaught error when handling `/proc/{pid}/`: {e}");
             }
+        }
+        // END pid loop
```
blt, line 129: This loop is now short enough you could cut these.

lading/src/observer/linux/procfs.rs
```diff
@@ -118,158 +118,205 @@ impl Sampler {
                     }
                 }
             }
+            // Update the process_info map to only hold processes seen by the current poll call.
+            self.process_info.retain(|pid, _| pids.contains(pid));
```
blt, line 122: This could potentially be a separate PR, since it's _technically_ out of scope for the change description.

blt (approved):

lading/src/observer/linux/procfs.rs
cmetz100: Since it fits within the spirit of handling pids that exit I added a line to the change description

PR #1154, opened 2024-12-16
Container Generator

scottopell:
### What does this PR do?

A form of load can be a container that exists and exposes some data. 

This generator aims to provide a way to start a generic container based on user's config.

### Motivation

Replace `checks-churn` benchmark from AML cluster

PR #1227, opened 2025-01-31
[CONTINT-4562] Container generator

L3n41c:
### What does this PR do?

A form of load can be a container that exists and exposes some data.

This generator aims to provide a way to start a generic container based on user's config.

### Motivation

Test the agent performances when it monitors many containers

### Additional Notes

Follow-up of #1154.

### DIFF
--- a/lading/src/generator.rs
+++ b/lading/src/generator.rs
@@ -15,6 +15,7 @@ use tracing::{error, warn};
 
 use crate::target::TargetPidReceiver;
 
+pub mod container;
 pub mod file_gen;
 pub mod file_tree;
 pub mod grpc;
@@ -67,6 +68,9 @@ pub enum Error {
     /// See [`crate::generator::procfs::Error`] for details.
     #[error(transparent)]
     ProcFs(#[from] procfs::Error),
+    /// See [`crate::generator::container::Error`] for details.
+    #[error(transparent)]
+    Container(#[from] container::Error),
 }
 
 #[derive(Debug, Deserialize, Serialize, PartialEq)]
@@ -120,6 +124,8 @@ pub enum Inner {
     ProcessTree(process_tree::Config),
     /// See [`crate::generator::procfs::Config`] for details.
     ProcFs(procfs::Config),
+    /// See [`crate::generator::container::Config`] for details.
+    Container(container::Config),
 }
 
 #[derive(Debug)]
@@ -152,6 +158,8 @@ pub enum Server {
     ProcessTree(process_tree::ProcessTree),
     /// See [`crate::generator::procfs::Procfs`] for details.
     ProcFs(procfs::ProcFs),
+    /// See [`crate::generator::container::Container`] for details.
+    Container(container::Container),
 }
 
 impl Server {
@@ -204,6 +212,9 @@ impl Server {
                 Self::ProcessTree(process_tree::ProcessTree::new(&conf, shutdown)?)
             }
             Inner::ProcFs(conf) => Self::ProcFs(procfs::ProcFs::new(&conf, shutdown)?),
+            Inner::Container(conf) => {
+                Self::Container(container::Container::new(config.general, &conf, shutdown)?)
+            }
         };
         Ok(srv)
     }
@@ -237,6 +248,8 @@ impl Server {
             Server::PassthruFile(inner) => inner.spin().await?,
             Server::ProcessTree(inner) => inner.spin().await?,
             Server::ProcFs(inner) => inner.spin().await?,
+            // Run the container generator
+            Server::Container(inner) => inner.spin().await?,
         };
 
         Ok(())
--- a/lading/src/generator/container.rs
+++ b/lading/src/generator/container.rs
@@ -1,0 +1,196 @@
+//! The container generator
+//!
+//! This generator is meant to spin up a container from a configured image. For now,
+//! it does not actually do anything beyond logging that it's running and then waiting
+//! for a shutdown signal.
+//!
+//! ## Future Work
+//! - Pull and run a specified container image
+//! - Possibly support metrics about container lifecycle
+
+use bollard::container::{
+    Config as ContainerConfig, CreateContainerOptions, RemoveContainerOptions,
+    StartContainerOptions, StopContainerOptions,
+};
+use bollard::image::CreateImageOptions;
+use bollard::secret::ContainerCreateResponse;
+use bollard::Docker;
+use serde::{Deserialize, Serialize};
+use tokio::pin;
+use tokio_stream::StreamExt;
+use tracing::{info, warn};
+use uuid::Uuid;
+
+use super::General;
+
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
+#[serde(deny_unknown_fields)]
+/// Configuration of the container generator.
+pub struct Config {
+    /// The seed for random operations (not currently used, but included for parity)
+    //pub seed: [u8; 32],
+    /// The container repository (e.g. "library/nginx")
+    pub repository: String,
+    /// The container image tag (e.g. "latest")
+    pub tag: String,
+    /// Arguments to provide to the container (docker calls this args, but that's a dumb name)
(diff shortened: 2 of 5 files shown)


blt (commented):
I really like where this is headed.

lading/src/generator/container.rs
```diff
@@ -0,0 +1,245 @@
...
+                    return Err(e.into());
+                }
+            }
+        }
+
+        let mut containers = Vec::with_capacity(self.nb);
+
+        for _ in 0..self.nb {
+            let container_name = format!("lading_container_{}", Uuid::new_v4());
+            info!("Creating container: {}", container_name);
```
blt, line 150: This could be slightly changed to `info!("Creating container: {container_name}");`. String formatting allows for variable names to be present within `{}`.

lading/src/generator/container.rs
```diff
@@ -0,0 +1,245 @@
...
+                                .iter()
+                                .map(|port| (port.as_str(), HashMap::new()))
+                                .collect()
+                        }),
+                        ..Default::default()
+                    },
+                )
+                .await?;
+
+            info!("Created container with id: {}", container.id);
```
blt, line 188: REF the above string formatting comment, if you wanted to do the same thing here you'd have to define `container.id` a name, so like `info!("Created container with id: {id}", id = container.id);` More an FYI than anything. It can be helpful when there are a lot of fields to keep track of.

lading/src/generator/container.rs
```diff
@@ -0,0 +1,245 @@
...
+    /// 3. Create and start the container.
+    /// 4. Wait for shutdown signal.
+    /// 5. On shutdown, stop and remove the container.
+    pub async fn spin(self) -> Result<(), Error> {
+        info!("Container generator running: {}:{}", self.image, self.tag);
+
+        let docker = Docker::connect_with_local_defaults()?;
+
+        let full_image = format!("{}:{}", self.image, self.tag);
+        info!("Ensuring image is available: {}", full_image);
```
blt, line 120: This and the below logs, could we bump these down to debug? In practice we'll see tons of these logs and they strike me as more useful in a debugging scenario. CC @cmetz100, I'm curious also for your thoughts. I know log volume in our system is an area you think on a fair bit.

lading/src/generator/container.rs
```diff
@@ -0,0 +1,245 @@
...
+/// Represents a container that can be spun up from a configured image.
+pub struct Container {
+    image: String,
+    tag: String,
+    args: Option<Vec<String>>,
+    env: Option<Vec<String>>,
+    labels: Option<HashMap<String, String>>,
+    network_disabled: Option<bool>,
+    exposed_ports: Option<Vec<String>>,
+    nb: usize,
```
blt, line 71: Ah yeah, following on from the comment on the Config field, `nb` as a field name here gives me very little understanding of what it is for.

lading/src/generator/container.rs
```diff
@@ -0,0 +1,245 @@
...
+    /// Environment variables to set in the container
+    pub env: Option<Vec<String>>,
+    /// Labels to apply to the container
+    pub labels: Option<HashMap<String, String>>,
+    /// Network mode to use for the container
+    pub network_disabled: Option<bool>,
+    /// Ports to expose from the container
+    pub exposed_ports: Option<Vec<String>>,
+    /// Number of containers to spin up (defaults to 1)
+    pub nb_containers: Option<u32>,
```
blt, line 47: Looking at this I'm not totally sure what `nb` means. I think probably "number"? In general lading house style is to spell things out, so probably `total_containers` and further adjustment of `nb` below is in keeping with that.

blt (commented):
Some style comments on the commits. I would like to clarify what happens if the spawned container exits. As I read it the `Container` generator will not check for spawned container liveness and so if a spawned container fails the user won't know until `stop_and_remove_container` is called? There's potentially a long period of time between startup and shutdown so an earlier warning -- of outright failure of lading -- is desirable. I could buy may-fail or must-not-fail approaches to this generator, but documenting which approach is taken here is helpful to the user.

lading/src/generator.rs
```diff
@@ -239,6 +250,8 @@ impl Server {
             Server::PassthruFile(inner) => inner.spin().await?,
             Server::ProcessTree(inner) => inner.spin().await?,
             Server::ProcFs(inner) => inner.spin().await?,
+            // Run the container generator
```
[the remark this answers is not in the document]
L3n41c: Agreed! Fixed in https://github.com/DataDog/lading/pull/1227/commits/59d040caefd33fe0de219df051dbfd89d3dce03c.

lading/src/generator/container.rs
```diff
@@ -0,0 +1,245 @@
...
+#[serde(deny_unknown_fields)]
+/// Configuration of the container generator.
+pub struct Config {
+    /// The seed for random operations (not currently used, but included for parity)
+    //pub seed: [u8; 32],
+    /// The container repository (e.g. "library/nginx")
+    pub repository: String,
+    /// The container image tag (e.g. "latest")
+    pub tag: String,
+    /// Arguments to provide to the container (docker calls this args, but that's a dumb name)
```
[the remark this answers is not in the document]
L3n41c: Agreed! Fixed in https://github.com/DataDog/lading/pull/1227/commits/59d040caefd33fe0de219df051dbfd89d3dce03c.

lading/src/generator/container.rs
L3n41c: Fixed in https://github.com/DataDog/lading/pull/1227/commits/59d040caefd33fe0de219df051dbfd89d3dce03c.

lading/src/generator/container.rs
L3n41c: 👍🏻

lading/src/generator/container.rs
L3n41c: I switched all the `info` logs to `debug` except the very first `Container generator running …` one in https://github.com/DataDog/lading/pull/1227/commits/59d040caefd33fe0de219df051dbfd89d3dce03c.

lading/src/generator/container.rs
L3n41c: I renamed it `number_of_containers` in https://github.com/DataDog/lading/pull/1227/commits/59d040caefd33fe0de219df051dbfd89d3dce03c.

lading/src/generator/container.rs
L3n41c: I renamed it `number_of_containers` in https://github.com/DataDog/lading/pull/1227/commits/59d040caefd33fe0de219df051dbfd89d3dce03c.

L3n41c:
@blt Thank you for your review!
I should have addressed all the comments.

> I would like to clarify what happens if the spawned container exits. As I read it the Container generator will not check for spawned container liveness and so if a spawned container fails the user won't know until stop_and_remove_container is called? There's potentially a long period of time between startup and shutdown so an earlier warning -- of outright failure of lading -- is desirable.

I’ve also updated the PR so that `lading` is now periodically checking if the spawned containers are still running. If they’re not, a warning is issued.

lading/src/generator/container.rs
```diff
@@ -0,0 +1,251 @@
...
+            debug!("Started container: {id}", id = container.id);
+        }
+
+        // Wait for shutdown signal
+        let shutdown_wait = self.shutdown.recv();
+        tokio::pin!(shutdown_wait);
+        loop {
+            tokio::select! {
+                // Check that containers are still running every 10 seconds
+                () = tokio::time::sleep(tokio::time::Duration::from_secs(10)) => {
```
blt, line 204: This would be better done in terms of `tokio::time::interval`, example [here](https://github.com/DataDog/lading/blob/68b7642e0c7dc59b69559aa43f5ddb1329cb0a57/lading/src/bin/lading.rs#L529).

blt (commented):
One small comment about using interval instead of repeatedly creating sleeps, otherwise looks good. I'll be happy to see this merge up after that's fixed.

blt (approved):

PR #1231, opened 2025-02-04
Ignore metrics read over prometheus with NaN value

cmetz100:
### What does this PR do?

Ignores counters and guages scraped over prometheus that have a NaN value.

### Motivation

Turns out NaN is a valid [constant](https://doc.rust-lang.org/std/f64/constant.NAN.html) of f64 so when we scrape prometheus endpoint and read a NaN value we try to set it as a gauge or counter. This is then translated to null in the capture file. null in the capture file causes `rouster-series-sender` to fail when submitting metrics

### Related issues

I am not 100% confident that this is all we need to handle all edges regarding NaN

### DIFF
--- a/lading/src/target_metrics/prometheus.rs
+++ b/lading/src/target_metrics/prometheus.rs
@@ -276,6 +276,11 @@ pub(crate) async fn scrape_metrics(
                     }
                 };
 
+                if value.is_nan() {
+                    warn!("Skipping NaN guage value");
+                    continue;
+                }
+
                 gauge!(format!("target/{name}"), &all_labels.unwrap_or_default()).set(value);
             }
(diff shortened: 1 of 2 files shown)


blt (approved):

lading/src/target_metrics/prometheus.rs
```diff
@@ -276,6 +276,11 @@ pub(crate) async fn scrape_metrics(
                     }
                 };
 
+                if value.is_nan() {
```
scottopell, line 279: quick skim of the rust std lib and `is_finite` sticks out to me as potentially useful, it covers both `NaN` and pos/neg infinity. 
https://doc.rust-lang.org/nightly/std/primitive.f64.html#method.is_infinite

PR #1257, opened 2025-02-25
Fix process-tree generator

vitkyrka:
### What does this PR do?

Fix the following errors which occur when using the process_tree generator (for example, with examples/lading.yaml):

 ERROR lading: Generator shut down unexpectedly: Execution error: execution failed:
 Command process-tree-gen: Argument group 'config' contains non-existent argument 'config-path'

 ERROR lading::generator::process_tree: process tree generator execution error
 ERROR lading: Generator shut down unexpectedly: Execution error:
  execution failed: error: the following required arguments were not
  provided:
   <--capture-path <CAPTURE_PATH>|...>

### Motivation

Error found while running example.

### DIFF
--- a/lading/src/generator/process_tree.rs
+++ b/lading/src/generator/process_tree.rs
@@ -320,10 +320,9 @@ impl ProcessTree {
         loop {
             tokio::select! {
                 _ = self.throttle.wait() => {
-                    // Dummy --capture-path and --target-pid just to pass laging clap constraints
                     let output = Command::new(lading_path)
-                        .args(["--capture-path", "dummy"])
-                        .args(["--target-pid", "1"])
+                        .args(["--no-capture"])
+                        .args(["--no-target"])
                         .arg("process-tree-gen")
                         .arg("--config-content")
                         .arg(&self.config_content)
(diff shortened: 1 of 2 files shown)


lading/src/generator/process_tree.rs
```diff
@@ -320,8 +320,9 @@ impl ProcessTree {
         loop {
             tokio::select! {
                 _ = self.throttle.wait() => {
-                    // using pid as target pid just to pass laging clap constraints
+                    // Dummy --capture-path and --target-pid just to pass laging clap constraints
                     let output = Command::new(lading_path)
+                        .args(["--capture-path", "dummy"])
```
blt, line 325: Hmm, will this not end up writing a `dummy` file? Seems cleaner to me to support a `--no-capture` option or similar.

lading/src/generator/process_tree.rs
```diff
@@ -320,8 +320,9 @@ impl ProcessTree {
...
                         .args(["--target-pid", "1"])
```
blt, line 326: I believe there's a flag to track no pid that would be appropriate to use here.

lading/src/generator/process_tree.rs
vitkyrka: Added `--no-capture` and used it here.

lading/src/generator/process_tree.rs
vitkyrka: Changed this to `--no-target`

scottopell:
this has been removed

PR #1302, opened 2025-03-27
Ignore forked but not exec'd child processes

blt:
### What does this PR do?

This commit introduces code into the procfs Sampler such that discovered
children that are not exec'd yet are not polled. We have found that in
rare cases this causes a double-count of the parent's memory, although it's
not accumulated as such by Linux itself.

### DIFF
--- a/lading/src/observer/linux/procfs.rs
+++ b/lading/src/observer/linux/procfs.rs
@@ -5,11 +5,11 @@ mod uptime;
 
 use std::{collections::VecDeque, io};
 
-use metrics::gauge;
+use metrics::{counter, gauge};
 use nix::errno::Errno;
 use procfs::process::Process;
 use rustc_hash::{FxHashMap, FxHashSet};
-use tracing::{error, warn};
+use tracing::{error, info, warn};
 
 const BYTES_PER_KIBIBYTE: u64 = 1024;
 
@@ -163,6 +163,8 @@ impl Sampler {
                                 if child_info.exe == parent_info.exe
                                     && child_info.cmdline == parent_info.cmdline
                                 {
+                                    counter!("process_skipped").increment(1);
+                                    info!("Skipping process {pid}: {child_info:?}");
                                     continue;
                                 }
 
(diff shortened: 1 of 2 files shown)


blt:
* **#1302** <a href="https://app.graphite.dev/github/pr/DataDog/lading/1302?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/lading/1302?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lading/src/observer/linux/procfs.rs
```diff
@@ -87,13 +84,36 @@ impl Sampler {
...
         // Every sample run we collect all the child processes rooted at the
         // parent. As noted by the procfs documentation is this done by
         // dereferencing the `/proc/<pid>/root` symlink.
-        let mut pids: FxHashSet<i32> = FxHashSet::default();
+        // We must be sure to initialize the parent process info.
+        let parent_pid = self.parent.pid();
+        let parent_info = match initialize_process_info(parent_pid).await {
+            Ok(Some(info)) => info,
+            Ok(None) => {
+                warn!("Could not initialize parent process info");
```
scottopell, line 99: I was going back and forth on whether this is fatal or not, and I agree that this is fine to return `Ok`, if we return an `Err`, then the observer server will crash and this will never retry, which is not the behavior we want. This is a retry-able error.

scottopell (approved):

preinlein (approved):

lading/src/observer/linux/procfs.rs
preinlein: To your point, this is an error and not a warning.

Technically `lading` recovers but an error has happened that results in unexpected results.

blt:
### Merge activity

* **Mar 31, 2:37 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/DataDog/lading/1302).

PR #1317, opened 2025-04-09
Generators obey throttle wait_for errors

blt:
### What does this PR do?

A brief description of the change being made with this pull request.

### Motivation

### Related issues

for this Pull Request.

### Additional Notes


GeorgeHahn (approved):
Very nice find. This code gives me a feeling that we have mismatch somewhere, but I can't quite put my finger on where. We chatted about having the throttle handle this situation in slack - that'd lead to bursty output and probably isn't what we'd want across the board. Suppressing blocks that are too large at the output seems like it'll shape the payloads we send in surprising ways, but I don't think the throttle should influence payload generation either. Anyways, I think there's something else here, but I have no idea what it is right now, and this change is definitely a step forward.

blt:
> Very nice find. This code gives me a feeling that we have mismatch somewhere, but I can't quite put my finger on where. We chatted about having the throttle handle this situation in slack - that'd lead to bursty output and probably isn't what we'd want across the board. Suppressing blocks that are too large at the output seems like it'll shape the payloads we send in surprising ways, but I don't think the throttle should influence payload generation either. Anyways, I think there's something else here, but I have no idea what it is right now, and this change is definitely a step forward.

I feel similarly. The consolation here is, I think, most users don't set throughput so low and if they do now it's plain that something's goofy. Yesterday I considered adjusting the maximum block size automatically based on the throughput, cap it at 1MiB as-is but allow for a 0.75*throughput default if that's needed. Now, hmm, I don't know about that idea.

blt:
### Merge activity

* **Apr 10, 5:42 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/DataDog/lading/1317).

PR #1322, opened 2025-04-16
[CONTINT-4562] Add WSS measurement (Working Set Size)

L3n41c:
### What does this PR do?

Implement WSS (Working Set Size) measurement.

### Motivation

We need a metric that measures how much memory the target workload really uses.
The metric we currently use to asses memory consumption, RSS and PSS, are more measuring the amount of memory that has been allocated and it includes reclaimable memory.

### Additional Notes

See https://www.brendangregg.com/wss.html

Here is how to validate the new `total_wss_bytes` metric reported value:

1. Start an agent with no memory limit with:
```console
$ docker run --rm --cgroupns host --pid host --name dd-agent -e DD_HOSTNAME=lenaic-Precision-5570 -v /var/run/docker.sock:/var/run/docker.sock:ro -v /proc/:/host/proc/:ro -v /sys/fs/cgroup/:/host/sys/fs/cgroup:ro -e DD_API_KEY=$DD_API_KEY gcr.io/datadoghq/agent:7.65.0-rc.9 bash -c 'ln -s /etc/datadog-agent/datadog-docker.yaml /etc/datadog-agent/datadog.yaml && exec /opt/datadog-agent/bin/agent/agent run'
```

2. Collect the memory metrics with `lading`:
```console
$ target/debug/lading --config-path examples/lading-idle.yaml --target-container dd-agent --prometheus-addr 127.0.0.1:8080 --warmup-duration-seconds 0 --experiment-duration-seconds 3600
```

```console
$ curl -s http://localhost:8080 | grep -E 'total_[prw]ss_bytes'
# TYPE total_rss_bytes gauge
total_rss_bytes 190119936
# TYPE total_wss_bytes gauge
total_wss_bytes 108711936
# TYPE total_pss_bytes gauge
total_pss_bytes 190111744
```

| metric | value in bytes |
| :-- | --: |
| `total_pss_bytes` | 190111744 |
| `total_rss_bytes` | 190119936 |
| `total_wss_bytes` | 108711936 |

3. Stop the agent and restart it with a memory limit equal to the measured WSS + 10% of safety margin (119 MiB):

```console
$ docker run --rm --cgroupns host --pid host --name dd-agent -e DD_HOSTNAME=lenaic-Precision-5570 -v /var/run/docker.sock:/var/run/docker.sock:ro -v /proc/:/host/proc/:ro -v /sys/fs/cgroup/:/host/sys/fs/cgroup:ro -e DD_API_KEY=$DD_API_KEY --memory 119m --memory-swap 119m gcr.io/datadoghq/agent:7.65.0-rc.9 bash -c 'ln -s /etc/datadog-agent/datadog-docker.yaml /etc/datadog-agent/datadog.yaml && exec /opt/datadog-agent/bin/agent/agent run'
```

Observe that the agent is stable (10 minutes of uptime with no restart):

```console
$ docker ps
CONTAINER ID   IMAGE                                COMMAND                  CREATED          STATUS                    PORTS                                                                    NAMES
27beb411fc21   gcr.io/datadoghq/agent:7.65.0-rc.9   "bash -c 'ln -s /etc…"   10 minutes ago   Up 10 minutes (healthy)   8125/udp, 8126/tcp                                                       dd-agent
```

5. Stop the agent and restart it with a memory limit equal to 10% less that WSS (98 MiB):

```console
$ docker run --rm --cgroupns host --pid host --name dd-agent -e DD_HOSTNAME=lenaic-Precision-5570 -v /var/run/docker.sock:/var/run/docker.sock:ro -v /proc/:/host/proc/:ro -v /sys/fs/cgroup/:/host/sys/fs/cgroup:ro -e DD_API_KEY=$DD_API_KEY --memory 98m --memory-swap 98m gcr.io/datadoghq/agent:7.65.0-rc.9 bash -c 'ln -s /etc/datadog-agent/datadog-docker.yaml /etc/datadog-agent/datadog.yaml && exec /opt/datadog-agent/bin/agent/agent run'
```

Observe that the agent is eventually OOM killed:
```
avril 16 15:24:44 lenaic-Precision-5570 kernel: Memory cgroup out of memory: Killed process 174731 (agent) total-vm:3400360kB, anon-rss:77848kB, file-rss:96800kB, shmem-rss:0kB, UID:0 pgta>
avril 16 15:24:44 lenaic-Precision-5570 kernel: agent invoked oom-killer: gfp_mask=0xcc0(GFP_KERNEL), order=0, oom_score_adj=0
```

So, the value reported by `total_wss_bytes` represents the boundary between a memory limit too low that triggers OOM kills and a memory limit high enough to let the target workload run fine.

### DIFF
--- a/lading/Cargo.toml
+++ b/lading/Cargo.toml
@@ -36,6 +36,7 @@ flate2 = { version = "1.1.1", default-features = false, features = [
 ] }
 futures = "0.3.31"
 fuser = { version = "0.15", optional = true }
+page_size = { version = "0.6.0" }
 heck  = { version = "0.5", default-features = false }
 http = { workspace = true }
 http-body-util = { workspace = true }
--- a/lading/src/observer/linux.rs
+++ b/lading/src/observer/linux.rs
@@ -1,5 +1,6 @@
 mod cgroup;
 mod procfs;
+mod wss;
 
 use tracing::error;
 
@@ -12,24 +13,32 @@ pub enum Error {
     /// Wrapper for [`procfs::Error`]
     #[error("Procfs: {0}")]
     Procfs(#[from] procfs::Error),
+    /// Wrapper for [`wss::Error`]
+    #[error("WSS: {0}")]
+    Wss(#[from] wss::Error),
 }
 
 #[derive(Debug)]
 pub(crate) struct Sampler {
     procfs: procfs::Sampler,
     cgroup: cgroup::Sampler,
+    wss: wss::Sampler,
     smaps_interval: u8,
+    last_wss_sample: tokio::time::Instant,
 }
 
 impl Sampler {
     pub(crate) fn new(parent_pid: i32, labels: Vec<(String, String)>) -> Result<Self, Error> {
         let procfs_sampler = procfs::Sampler::new(parent_pid)?;
         let cgroup_sampler = cgroup::Sampler::new(parent_pid, labels)?;
+        let wss_sampler = wss::Sampler::new(parent_pid)?;
 
         Ok(Self {
             procfs: procfs_sampler,
             cgroup: cgroup_sampler,
+            wss: wss_sampler,
             smaps_interval: 10,
+            last_wss_sample: tokio::time::Instant::now() - tokio::time::Duration::from_secs(61),
         })
     }
 
@@ -44,6 +53,14 @@ impl Sampler {
 
         self.procfs.poll(sample_smaps).await?;
         self.cgroup.poll().await?;
+
+        // WSS measures the amount of memory that has been accessed since the last poll.
+        // As a consequence, the poll interval impacts the measure.
+        // That’s why we need to be sure we don’t poll more often than once per minute.
+        if self.last_wss_sample.elapsed() > tokio::time::Duration::from_secs(60) {
+            self.wss.poll().await?;
+            self.last_wss_sample = tokio::time::Instant::now();
+        }
 
         Ok(())
     }
--- a/lading/src/observer/linux/wss.rs
+++ b/lading/src/observer/linux/wss.rs
@@ -1,0 +1,214 @@
+use metrics::gauge;
+use procfs::process::{MemoryMap, MemoryPageFlags, PageInfo, Pfn, Process};
+use std::collections::HashMap;
+use std::io::{Read, Seek, SeekFrom, Write};
+use tracing::debug;
+
+const PAGE_OFFSET: u64 = 0xffff_8800_0000_0000;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    /// Wrapper for [`std::io::Error`]
+    #[error("IO error: {0}")]
+    Io(#[from] std::io::Error),
+    /// Wrapper for [`procfs::ProcError`]
+    #[error("Unable to read procfs: {0}")]
+    Proc(#[from] procfs::ProcError),
+}
+
+#[derive(Debug)]
+pub(crate) struct Sampler {
+    parent_pid: i32,
+}
+
+impl Sampler {
+    pub(crate) fn new(parent_pid: i32) -> Result<Self, Error> {
+        Ok(Self { parent_pid })
+    }
+
+    pub(crate) async fn poll(&mut self) -> Result<(), Error> {
+        let page_size = page_size::get();
+        let mut pfn_set = PfnSet::new();
+
+        for process in ProcessDescendentsIterator::new(self.parent_pid) {
+            debug!("Process PID: {}", process.pid());
+            let mut pagemap = process.pagemap()?;
+            for MemoryMap {
+                address: (begin, end),
+                ..
+            } in process.maps()?
+            {
+                if begin > PAGE_OFFSET {
+                    continue; // page idle tracking is user mem only
+                }
+                debug!("Memory region: {:#x} — {:#x}", begin, end);
+                let begin = begin as usize / page_size;
+                let end = end as usize / page_size;
+                for page in pagemap.get_range_info(begin..end)? {
+                    if let PageInfo::MemoryPage(memory_page_flags) = page {
+                        if memory_page_flags.contains(MemoryPageFlags::PRESENT) {
+                            pfn_set.insert(memory_page_flags.get_page_frame_number());
+                        }
+                    }
+                }
+            }
+        }
+
+        let mut nb_pages = 0;
+
+        // See https://www.kernel.org/doc/html/latest/admin-guide/mm/idle_page_tracking.html
+        let mut page_idle_bitmap = std::fs::OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open("/sys/kernel/mm/page_idle/bitmap")?;
+
+        for (pfn_block, pfn_bitset) in pfn_set {
+            page_idle_bitmap.seek(SeekFrom::Start(pfn_block * 8))?;
+
+            let mut buffer = [0; 8];
+            page_idle_bitmap.read_exact(&mut buffer)?;
+            let bitset = u64::from_ne_bytes(buffer);
+
+            nb_pages += (!bitset & pfn_bitset).count_ones() as usize;
+
+            page_idle_bitmap.seek(SeekFrom::Start(pfn_block * 8))?;
+            page_idle_bitmap.write_all(&pfn_bitset.to_ne_bytes())?;
+        }
+
+        gauge!("total_wss_bytes").set((nb_pages * page_size) as f64);
+
+        Ok(())
+    }
+}
+
+struct ProcessDescendentsIterator {
+    stack: Vec<Process>,
+}
+
+impl ProcessDescendentsIterator {
+    fn new(parent_pid: i32) -> Self {
+        Self {
+            stack: vec![
+                Process::new(parent_pid).expect(format!("process {parent_pid} not found").as_str()),
+            ],
+        }
+    }
+}
+
+impl Iterator for ProcessDescendentsIterator {
+    type Item = Process;
+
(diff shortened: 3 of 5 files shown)


lading/src/observer/linux.rs
```diff
@@ -45,6 +62,16 @@ impl Sampler {
         self.procfs.poll(sample_smaps).await?;
         self.cgroup.poll().await?;
 
+        if let Some(wss) = &mut self.wss {
+            // WSS measures the amount of memory that has been accessed since the last poll.
+            // As a consequence, the poll interval impacts the measure.
+            // That’s why we need to be sure we don’t poll more often than once per minute.
+            if self.last_wss_sample.elapsed() > tokio::time::Duration::from_secs(60) {
```
scottopell, line 69: How do we arrive at 60s as a good poll interval?

scottopell, line 69: Separately, we already sample `smaps` access every 10s, logic for this is directly above https://github.com/DataDog/lading/blob/main/lading/src/observer/linux.rs#L36-L49

It would be nice if we could line up these sampling intervals so that we can correlate if needed.

lading/src/observer/linux/wss.rs
```diff
@@ -0,0 +1,99 @@
...
+use tracing::debug;
+
+mod process_descendents;
+use process_descendents::ProcessDescendantsIterator;
+
+mod pfnset;
+use pfnset::PfnSet;
+
+pub(super) const PAGE_IDLE_BITMAP: &str = "/sys/kernel/mm/page_idle/bitmap";
+// From https://github.com/torvalds/linux/blob/c62f4b82d57155f35befb5c8bbae176614b87623/arch/x86/include/asm/page_64_types.h#L42
```
scottopell, line 14: is this amd64 specific? If so, lets represent that here in the code so that we can add support for aarch64 in the future if needed.

lading/src/observer/linux.rs
```diff
@@ -12,24 +13,40 @@ pub enum Error {
...
 
 impl Sampler {
     pub(crate) fn new(parent_pid: i32, labels: Vec<(String, String)>) -> Result<Self, Error> {
         let procfs_sampler = procfs::Sampler::new(parent_pid)?;
         let cgroup_sampler = cgroup::Sampler::new(parent_pid, labels)?;
+        let wss_sampler = if wss::Sampler::is_available() {
+            Some(wss::Sampler::new(parent_pid)?)
+        } else {
+            warn!(
+                "{} isn’t accessible. Either the kernel hasn’t been compiled with CONFIG_IDLE_PAGE_TRACKING or the process doesn’t have access to it. WSS sampling is not available",
```
scottopell, line 38: Is there a specific capability that we can include in this error message as a hint for providing access?

lading/src/observer/linux/wss.rs
```diff
@@ -0,0 +1,99 @@
...
+                .open(PAGE_IDLE_BITMAP)?,
+        })
+    }
+
+    #[allow(clippy::unused_async)]
+    pub(crate) async fn poll(&mut self) -> Result<(), Error> {
+        let page_size = page_size::get();
+        let mut pfn_set = PfnSet::new();
+
+        for process in ProcessDescendantsIterator::new(self.parent_pid) {
```
scottopell, line 54: This seems potentially useful in our `procfs` iteration code as well, I'd like to only have a single way to iterate through child processes ideally.

scottopell (commented):
Initial pass

lading/src/observer/linux.rs
L3n41c: The thing is that the agent doesn’t have an homogeneous workload that would execute the same pieces of code on the same pieces of data at all points in time.
Instead, it has a lot of periodic tasks that are scheduled every so often.
For ex., some python checks are scheduled once every 15s, the forwarder sends the aggregated data every 5s, the workloadmeta store polls the kubelet every 15s, etc.
So, if we were collecting WSS every second, the data accessed by the workloadmeta kubelet collector would be taken into account only in one sample out of 15, resulting in a very spiky metric.
In order to have a stable metric, we need a polling interval that is long enough to be sure that all the periodic tasks executed every so often by the agent have been executed during that interval.

With all the agent tasks executed every 15s, a WSS polling interval of 20s could theoretically have been enough. But experiments shown that it wasn’t enough to get a stable metric.

A possible future improvement (but in a follow up PR), could be to collect WSS at different intervals.

lading/Cargo.toml
```diff
@@ -36,6 +36,7 @@ flate2 = { version = "1.1.1", default-features = false, features = [
 ] }
 futures = "0.3.31"
 fuser = { version = "0.15", optional = true }
+page_size = { version = "0.6.0" }
```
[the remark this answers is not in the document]
L3n41c: I agree that the feature provided by this crate is so simple that we can copy it and get rid of this direct dependency.
That’s what I did in https://github.com/DataDog/lading/pull/1322/commits/c6fc25a50740dc48abf8ba4047dc8edae42e1813#diff-12dfd5a93fda4762b36064bede36873a0a6a1c3843a8f3167dc31ab915abaa36R63

However, please note that it doesn’t remove the dependency to `page_size` as this crate was already an indirect dependency prior to this PR as it can be seen on current `main` branch:
* https://github.com/DataDog/lading/blob/54558640f839c9cb39b4876fae8032e0d23b9144/Cargo.lock#L888
* https://github.com/DataDog/lading/blob/54558640f839c9cb39b4876fae8032e0d23b9144/Cargo.lock#L2148-L2156

lading/src/observer/linux.rs
```diff
@@ -12,24 +13,40 @@ pub enum Error {
...
+            );
+            None
+        };
 
         Ok(Self {
             procfs: procfs_sampler,
             cgroup: cgroup_sampler,
+            wss: wss_sampler,
             smaps_interval: 10,
+            last_wss_sample: tokio::time::Instant::now() - tokio::time::Duration::from_secs(61),
```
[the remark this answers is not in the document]
L3n41c: I’ve refactored this logic to make identical to the one currently used by `smaps` polling in https://github.com/DataDog/lading/pull/1322/commits/c6fc25a50740dc48abf8ba4047dc8edae42e1813#diff-d0609dafc88a691f93eb0c7eaf235015ffc74b1bfbe98e3ca36aa149b5b1f0d1R68-R69.

lading/src/observer/linux.rs
L3n41c: I’ve refactored this logic to make it identical to the one currently used by `smaps` polling in https://github.com/DataDog/lading/pull/1322/commits/c6fc25a50740dc48abf8ba4047dc8edae42e1813#diff-d0609dafc88a691f93eb0c7eaf235015ffc74b1bfbe98e3ca36aa149b5b1f0d1R68-R69.

lading/src/observer/linux/wss.rs
L3n41c: I’ve added an architecture attribute in https://github.com/DataDog/lading/pull/1322/commits/c6fc25a50740dc48abf8ba4047dc8edae42e1813#diff-12dfd5a93fda4762b36064bede36873a0a6a1c3843a8f3167dc31ab915abaa36R14.

lading/src/observer/linux.rs
L3n41c: I’ve added some hints about what would need to be looked at in the warning log in https://github.com/DataDog/lading/pull/1322/commits/c6fc25a50740dc48abf8ba4047dc8edae42e1813#diff-d0609dafc88a691f93eb0c7eaf235015ffc74b1bfbe98e3ca36aa149b5b1f0d1R38-R52.

lading/src/observer/linux/wss.rs
L3n41c: I’ve updated the `procfs` observer to make it use the same `ProcessDescendantsIterator` in https://github.com/DataDog/lading/pull/1322/commits/c6fc25a50740dc48abf8ba4047dc8edae42e1813#diff-d8cfa4d9cb3461bf7decb34c7751bf84cc6fa811bf34d9100370c13eca3f500fR93-R105.

lading/src/observer/linux/wss/pfnset.rs
```diff
@@ -0,0 +1,47 @@
...
+
+/// Structure used to represent a set of page frame numbers (PFNs)
+/// in a format suitable for use with the Idle Page Tracking API.
+/// <https://www.kernel.org/doc/html/latest/admin-guide/mm/idle_page_tracking.html>
+#[derive(Debug)]
+pub(super) struct PfnSet(HashMap<u64, u64>);
+
+impl PfnSet {
+    pub(super) fn new() -> Self {
+        Self(HashMap::with_capacity(1024))
```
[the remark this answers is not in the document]
L3n41c: It’s always difficult to find the right balance between a too small value, that will lead to many re-allocations, and a too big value, that will lead to a waste of resources (not only memory, but also CPU as iterating over a HashMap is O(capacity) instead of O(len))

In this case, a HashMap item represents an 8-bytes block, each bit of which represents a page of 4 KiB of memory.
So, a HashMap item represents between 1 and 64 pages. I.e. between 4 KiB and 256 KiB of memory.
So, **a HashMap of 1024 items can represent between 1024×1×4 KiB = 4 MiB and 1024×64×4 KiB = 256 MiB** of memory.

Given the fact that the probability for a process to have its virtual memory mapped to a contiguous portion of physical memory is rather low, we’re most probably closer to the lower bound than to the upper one.
In practice, what I observed on my machine is around 3 bits set per block in average.

As the workload that will be monitored by lading usually consumes more few megabytes, this initial capacity is in fact rather low.

blt (approved):

PR #1325, opened 2025-04-22
[CONTINT-4562] Add a parameter to continuously stop and recreate containers

L3n41c:
### What does this PR do?

Add a new `max_lifetime` parameter to the `container` generator.
This will make `lading` endlessly delete the oldest container and replace it with a newer one.

### Motivation

Some memory leaks inside the agent might happen each time a new container is detected or each time an old container is deleted.
In order to catch them, having a a static set of running containers during the duration of an experiment might not be enough.
We would like to generate a continuous stream of container deletion and container re-creation events.

### Additional Notes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ## Added
+- Compute the Working Set Size of the target thanks to the Linux Idle Page Tracking API.
+  And expose it in the new `total_wss_bytes` metric.
 - Add a `max_lifetime` parameter to the `container` generator to make it generate
   a continuous stream of docker containers deletion and re-creation.
 
@@ -13,8 +15,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Added
 - Introduce a `container` generator able to generate an arbitrary number
   of docker containers.
-- Compute the Working Set Size of the target thanks to the Linux Idle Page Tracking API.
-  And expose it in the new `total_wss_bytes` metric.
 ## Changed
 - Lading's byte-unit crate is now updated to 5.x. This version of the crate is
   very strict about the difference between MiB, Mb etc.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+## Added
+- Add a `max_lifetime` parameter to the `container` generator to make it generate
+  a continuous stream of docker containers deletion and re-creation.
 
 ## [0.25.9]
 ## Added
 - Introduce a `container` generator able to generate an arbitrary number
-  of docker containers
+  of docker containers.
 ## Changed
 - Lading's byte-unit crate is now updated to 5.x. This version of the crate is
   very strict about the difference between MiB, Mb etc.
--- a/examples/lading-container.yaml
+++ b/examples/lading-container.yaml
@@ -19,4 +19,5 @@ generator:
       network_disabled: false
       exposed_ports:
         - 5000/tcp
+      max_lifetime: 20
       number_of_containers: 10
--- a/examples/lading-container.yaml
+++ b/examples/lading-container.yaml
@@ -19,5 +19,5 @@ generator:
       network_disabled: false
       exposed_ports:
         - 5000/tcp
-      max_lifetime: 20
+      max_lifetime_seconds: 20
       number_of_containers: 10
--- a/lading/src/generator/container.rs
+++ b/lading/src/generator/container.rs
@@ -129,15 +129,13 @@ impl Container {
         }
 
         let number_of_containers = self.config.number_of_containers.unwrap_or(1);
-        let mut containers = VecDeque::with_capacity(number_of_containers);
-
-        for _ in 0..number_of_containers {
-            containers.push_back(
-                self.config
-                    .create_and_start_container(&docker, &full_image)
-                    .await?,
-            );
-        }
+        let mut containers = futures::future::join_all(
+            (0..number_of_containers)
+                .map(|_| self.config.create_and_start_container(&docker, &full_image)),
+        )
+        .await
+        .into_iter()
+        .collect::<Result<VecDeque<_>, Error>>()?;
 
         // Wait for shutdown signal
         let shutdown_wait = self.shutdown.recv();
--- a/lading/src/generator/container.rs
+++ b/lading/src/generator/container.rs
@@ -13,6 +13,7 @@ use bollard::image::CreateImageOptions;
 use bollard::secret::ContainerCreateResponse;
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, VecDeque};
+use tokio::sync::mpsc;
 use tokio_stream::StreamExt;
 use tracing::{debug, info, warn};
 use uuid::Uuid;
@@ -52,6 +53,9 @@ pub enum Error {
     /// Error produced by the Bollard Docker client.
     #[error("Bollard/Docker error: {0}")]
     Bollard(#[from] bollard::errors::Error),
+    /// Error produced by tokio
+    #[error("Tokio error: {0}")]
+    Tokio(#[from] mpsc::error::SendError<ContainerCreateResponse>),
 }
 
 #[derive(Debug)]
@@ -140,6 +144,7 @@ impl Container {
         // Wait for shutdown signal
         let shutdown_wait = self.shutdown.recv();
         tokio::pin!(shutdown_wait);
+        let (new_container_tx, mut new_container_rx) = mpsc::channel(32);
         let mut recreate_interval = self.config.max_lifetime.map(|max_lifetime| {
             tokio::time::interval(tokio::time::Duration::from_millis(
                 1_000 * max_lifetime / number_of_containers as u64,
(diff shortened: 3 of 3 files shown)


examples/lading-container.yaml
```diff
@@ -19,4 +19,5 @@ generator:
       network_disabled: false
       exposed_ports:
         - 5000/tcp
+      max_lifetime: 20
```
scottopell, line 22:
```suggestion
      max_lifetime_seconds: 20
```

would prefer to be explicit so the units are clear

lading/src/generator/container.rs
```diff
@@ -37,8 +38,10 @@ pub struct Config {
     pub network_disabled: Option<bool>,
     /// Ports to expose from the container
     pub exposed_ports: Option<Vec<String>>,
+    /// Maximum lifetime of containers before being replaced
+    pub max_lifetime: Option<u64>,
     /// Number of containers to spin up (defaults to 1)
-    pub number_of_containers: Option<u32>,
+    pub number_of_containers: Option<usize>,
```
scottopell, line 44: `u32` makes more sense to me here, this isn't something that would vary with CPU architecture.

scottopell (commented):
Looks good overall, only had time for a brief pass

lading/src/generator/container.rs
blt: +1

lading/src/generator/container.rs
```diff
@@ -169,46 +132,51 @@ impl Container {
...
+        .into_iter()
+        .collect::<Result<VecDeque<_>, Error>>()?;
 
         // Wait for shutdown signal
         let shutdown_wait = self.shutdown.recv();
         tokio::pin!(shutdown_wait);
-        let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(10));
+        let (new_container_tx, mut new_container_rx) = mpsc::channel(32);
+        let mut recreate_interval = self.config.max_lifetime.map(|max_lifetime| {
+            tokio::time::interval(tokio::time::Duration::from_millis(
```
blt, line 149: If we're needing to multiple here by 1_000 it's not clear to me that `from_millis` is the correct `from_`.

lading/src/generator/container.rs
```diff
@@ -169,46 +132,51 @@ impl Container {
...
+                1_000 * max_lifetime / number_of_containers as u64,
+            ))
+        });
+        let mut liveness_interval = tokio::time::interval(tokio::time::Duration::from_secs(10));
         loop {
             tokio::select! {
+                // Destroy and replace containers
+                _ = if let Some(ref mut interval) = recreate_interval { interval.tick() } else { std::future::pending().await } => {
```
blt, line 157: `recreate_interval { interval.tick() } else { std::future::pending().await }`

I'm confused by this. What's this construct doing?

lading/src/generator/container.rs
```diff
@@ -220,42 +188,112 @@ impl Container {
...
+
+                // Shutdown
                 () = &mut shutdown_wait => {
                     debug!("shutdown signal received");
-                    for container in &containers {
-                        Self::stop_and_remove_container(&docker, container).await?;
-                    }
+                    drop(new_container_tx);
+
+                    futures::future::join_all(
```
blt, line 197: Instead of using the `futures` crate directly, please use `tokio`'s machinery. In this case you'd want a [JoinSet](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html).

lading/src/generator/container.rs
goxberry: `core::time::Duration` implements `Div<u32>` and does the right thing [internally](https://doc.rust-lang.org/nightly/src/core/time.rs.html#800-817), so `core::time::Duration::from_secs` seems more appropriate here.

lading/src/generator/container.rs
```diff
@@ -169,46 +132,51 @@ impl Container {
...
+        .collect::<Result<VecDeque<_>, Error>>()?;
 
         // Wait for shutdown signal
         let shutdown_wait = self.shutdown.recv();
         tokio::pin!(shutdown_wait);
-        let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(10));
+        let (new_container_tx, mut new_container_rx) = mpsc::channel(32);
+        let mut recreate_interval = self.config.max_lifetime.map(|max_lifetime| {
+            tokio::time::interval(tokio::time::Duration::from_millis(
+                1_000 * max_lifetime / number_of_containers as u64,
```
goxberry, line 150: Forbidding `number_of_containers == 0` would avoid a panic here.

lading/src/generator/container.rs
goxberry: The first tick of a `tokio::time::Interval` [returns immediately](https://docs.rs/tokio/latest/tokio/time/struct.Interval.html#examples), so the select will choose among the first `recreate_interval` tick and the first `liveness_interval` tick arbitrarily. Is that behavior intended?

L3n41c: @blt The user may or may not want the containers to be continuously re-created.
That’s why the `max_lifetime_seconds` parameter is an `Option<NonZero<u64>>`.
This `Option`al integer is turned into an `recreate_interval: Option<Interval>` few lines above.

So, now, I’m looking for a way to express in the `tokio::select!` block the fact that I want an arm to be triggered by `interval.tick()` if `Some(interval) = recreate_interval`.
But if `None = recreate_interval`, then I want this arm to be disabled.

Basically, I *think* that what I need in the `else` block is something that has the same type as `interval.tick()` but that never fires.

What would be the most idiomatic way to have a `tokio::select!` arm triggered by an `Option<Interval>` ?

@goxberry I think it doesn’t matter which one of the `recreate_interval` tick or the `liveness_interval` one is executed first. Will the other be properly executed at the next `loop` iteration?

lading/src/generator/container.rs
L3n41c: Thanks for the tip!
I’ve migrated to `JoinSet` in https://github.com/DataDog/lading/pull/1325/commits/fea93299c54706d7b6aac3303ed9c80d793f00b0#diff-884b2b115b4cc2ae969b3beed17e0ea31dea56389e3948697694c795ebc7eb73R144-R160.

lading/src/generator/container.rs
L3n41c: I’ve made `number_of_containers` a `NonZero` in https://github.com/DataDog/lading/pull/1325/commits/fea93299c54706d7b6aac3303ed9c80d793f00b0#diff-884b2b115b4cc2ae969b3beed17e0ea31dea56389e3948697694c795ebc7eb73R45.

lading/src/generator/container.rs
L3n41c: Fixed in https://github.com/DataDog/lading/pull/1325/commits/fea93299c54706d7b6aac3303ed9c80d793f00b0#diff-884b2b115b4cc2ae969b3beed17e0ea31dea56389e3948697694c795ebc7eb73R171-R172.

examples/lading-container.yaml
L3n41c: Fixed in https://github.com/DataDog/lading/pull/1325/commits/fea93299c54706d7b6aac3303ed9c80d793f00b0#diff-884b2b115b4cc2ae969b3beed17e0ea31dea56389e3948697694c795ebc7eb73R43.

scottopell (approved):
looks good to me pending @blt 's approval on the select interval bit.

lading/src/generator/container.rs
L3n41c: Type updated to `std::num::NonZeroU32` in https://github.com/DataDog/lading/pull/1325/commits/fea93299c54706d7b6aac3303ed9c80d793f00b0#diff-884b2b115b4cc2ae969b3beed17e0ea31dea56389e3948697694c795ebc7eb73R45.

lading/src/generator/container.rs
blt: > What would be the most idiomatic way to have a tokio::select! arm triggered by an Option<Interval> ?

Given the context, I think what you have make sense. I may tinker with this area after this merges up for my own curiosity, but I don't have an offhand more idiomatic method. 

> Will the other be properly executed at the next loop iteration?

It will. So long as it doesn't matter which one is executed first, this will work as you expect.

blt (approved):

CHANGELOG.md
```diff
@@ -5,11 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+- Add a `max_lifetime` parameter to the `container` generator to make it generate
+  a continuous stream of docker containers deletion and re-creation.
 
 ## [0.25.9]
 ## Added
 - Introduce a `container` generator able to generate an arbitrary number
-  of docker containers
+  of docker containers.
 - Compute the Working Set Size of the target thanks to the Linux Idle Page Tracking API.
   And expose it in the new `total_wss_bytes` metric.
```
L3n41c, line 17:
```suggestion
```

CHANGELOG.md
```diff
@@ -5,11 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+## Added
+- Add a `max_lifetime` parameter to the `container` generator to make it generate
+  a continuous stream of docker containers deletion and re-creation.
```
L3n41c, line 10:
```suggestion
## [Unreleased]
## Added
- Compute the Working Set Size of the target thanks to the Linux Idle Page Tracking API.
  And expose it in the new `total_wss_bytes` metric.
- Add a `max_lifetime` parameter to the `container` generator to make it generate
  a continuous stream of docker containers deletion and re-creation.
```

PR #1372, opened 2025-05-29
Initial creation of netflow v5 payload generation

scottopell:
### What does this PR do?

Implements netflow v5 records as a new payload type that can be sent over UDP

### Motivation

Load testing for netflow processors.


blt:
Closing, still interested in it but I want to keep the backlog focused on active work.

PR #1376, opened 2025-06-02
Update bollard dependency

blt:
### What does this PR do?

This commit is an attempt to demonstrate the claude CLI tool for
    the purposes of mostly mechanical dependency upgrades. This kind
    of work is a non-trivial effort sink for lading and other SMP
    projects, it'd be swell if a machine could do it.

I find the result here satisfying. I acted as critical reviewer,
    claude as implementor.

### DIFF
--- a/lading/src/generator/container.rs
+++ b/lading/src/generator/container.rs
@@ -5,11 +5,15 @@
 //! for a shutdown signal.
 
 use bollard::Docker;
-use bollard::container::{
-    Config as ContainerConfig, CreateContainerOptions, RemoveContainerOptions,
-    StartContainerOptions, StopContainerOptions,
+use bollard::models::ContainerCreateBody;
+use bollard::query_parameters::{
+    CreateContainerOptionsBuilder,
+    CreateImageOptionsBuilder,
+    InspectContainerOptionsBuilder,
+    RemoveContainerOptionsBuilder,
+    StartContainerOptions,
+    StopContainerOptionsBuilder,
 };
-use bollard::image::CreateImageOptions;
 use bollard::secret::ContainerCreateResponse;
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, VecDeque};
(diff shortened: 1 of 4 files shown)


lading/src/target.rs
```diff
@@ -209,7 +210,11 @@ impl Server {
         let docker = Docker::connect_with_socket_defaults()?;
 
         let pid: i64 = loop {
-            if let Ok(container) = docker.inspect_container(&config.name, None).await {
+            let inspect_options = InspectContainerOptionsBuilder::default().build();
```
GeorgeHahn, line 213: This looks unnecessary - could the second parameter below stay `None`?

lading/src/generator/container.rs
```diff
@@ -241,32 +239,35 @@ async fn pull_image(docker: &Docker, full_image: &str) -> Result<(), Error> {
 impl Config {
     /// Convert the `Container` instance to a `ContainerConfig` for the Docker API.
     #[must_use]
-    fn to_container_config<'a>(&'a self, full_image: &'a str) -> ContainerConfig<&'a str> {
-        ContainerConfig {
-            image: Some(full_image),
+    fn to_container_config(&self, full_image: &str) -> ContainerCreateBody {
```
GeorgeHahn, line 242: Note, changes in this fn look unrelated, but I'm not sad about them

lading/src/generator/container.rs
```diff
@@ -178,7 +178,8 @@ impl Container {
                 // Check that containers are still running every 10 seconds
                 _ = liveness_interval.tick() => {
                     for container in &containers {
-                        if let Some(state) = docker.inspect_container(&container.id, None).await?.state {
+                        let inspect_options = InspectContainerOptionsBuilder::default().build();
```
GeorgeHahn, line 181: Nit, building a default value (same as below)

GeorgeHahn (approved):

lading/src/target.rs
blt: I think that it could. I dinged this myself at first and then after further consideration felt that it added more detail I found useful while skimming.

lading/src/generator/container.rs
blt: Same, same.

PR #1422, opened 2025-07-14
Add `/proc/vmstat` observer

GeorgeHahn:
### What does this PR do?

Observe `/proc/vmstat` during runs.

### Motivation

This will give us host-level [information](https://docs.kernel.org/admin-guide/mm/transhuge.html#monitoring-usage) about transparent hugepages.

### DIFF
--- a/lading/src/observer/linux/procfs/vmstat.rs
+++ b/lading/src/observer/linux/procfs/vmstat.rs
@@ -90,7 +90,7 @@ mod test {
 
     #[test]
     fn parse_vmstat_malformed_line() {
-        let vmstat_content = "nr_free_pages 123456\ninvalid line\n";
+        let vmstat_content = "nr_free_pages 123456\ninvalid_line\n";
         let result = proc_vmstat_inner(vmstat_content);
 
         assert!(result.is_err());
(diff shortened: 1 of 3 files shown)


lading/src/observer/linux/procfs/vmstat.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// This function reads the kernel's virtual memory statistics and publishes
+/// them as metrics. The vmstat file contains key-value pairs separated by
+/// whitespace, providing insights into memory usage, allocation, and pressure.
+pub(crate) async fn poll() -> Result<(), Error> {
+    let buf = tokio::fs::read_to_string("/proc/vmstat").await?;
+    let vmstat_data = proc_vmstat_inner(&buf)?;
+
+    // Publish metrics for all vmstat fields
+    for (field, value) in vmstat_data {
```
blt, line 28: If it were possible to avoid allocating in the inner function, considering this loop would be amenable to an iterator, that'd be cool. But, happy to have this in a follow-on PR.

blt (approved):

PR #1537, opened 2025-09-19
Try capture expiration another way

GeorgeHahn:
### What does this PR do?

Implement capture store with a map that is cleared on each iteration.

### Motivation

Metrics expiration, without deadlocks.


GeorgeHahn:
Upon discussion, this isn't quite right, but it gets towards something we would consider.

A complete change would use:
- A pair of ringbuffers, one used as the producer and one used as the consumer
- Writers must write through an atomic pointer
- The ringbuffers should be swapped at each interval

This allows:
- Metrics to be written in a nonblocking manner
- `metrics-rs` handles must work correctly

PR #1853, opened 2026-04-07
feat(otlp-metrics): added support for histogram, exp. histogram, and summary

lucastemb:
…m, and summary metrics

### What does this PR do?

Adds support for the remaining OTLP metrics (Exponential Histogram, Histogram, and Summary) in the correctness tests. 

### Motivation

Will be used to test whether these metric types are being properly handled by ADP in the Saluki repo when compared to the Agent. 

**Supports:**
- #https://github.com/DataDog/saluki/pull/1319
### Related issues

for this Pull Request.

### Additional Notes

Listed below is some additional information for the metrics added (how they are constructed and how they are expected to behave): 

- [Histogram](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram)
- [Exponential Histogram](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#exponentialhistogram)
- [Summary](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#summary-legacy)

### DIFF
--- a/lading_payload/src/opentelemetry/metric.rs
+++ b/lading_payload/src/opentelemetry/metric.rs
@@ -41,6 +41,8 @@
 pub(crate) mod templates;
 pub(crate) mod unit;
 
+use std::collections::BTreeMap;
+use std::hash::{DefaultHasher, Hash, Hasher};
 use std::rc::Rc;
 use std::{cell::RefCell, io::Write};
 
@@ -49,7 +51,11 @@ use crate::opentelemetry::common::templates::PoolError;
 use crate::{Error, common::config::ConfRange, common::strings};
 use bytes::BytesMut;
 use opentelemetry_proto::tonic::collector::metrics::v1::ExportMetricsServiceRequest;
-use opentelemetry_proto::tonic::metrics::v1::{ResourceMetrics, metric::Data, number_data_point};
+use opentelemetry_proto::tonic::common::v1::{KeyValue, any_value};
+use opentelemetry_proto::tonic::metrics::v1::{
+    Exemplar, Metric, ResourceMetrics, exemplar, exponential_histogram_data_point, metric::Data,
+    number_data_point,
+};
 use prost::Message;
 use serde::Deserialize;
 use templates::{Pool, ResourceTemplateGenerator};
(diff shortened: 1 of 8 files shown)


lading_payload/src/opentelemetry/metric.rs
```diff
@@ -403,29 +922,231 @@ impl<'a> SizedGenerator<'a> for OpentelemetryMetrics {
...
+                                    point.max = state.cumulative_max;
+                                } else {
+                                    point.bucket_counts = interval_bucket_counts;
+                                    point.count = interval_population.count;
+                                    point.sum = Some(interval_population.sum);
+                                    point.min = interval_population.min;
+                                    point.max = interval_population.max;
+                                }
+
+                                state.clock.note_emission(current_time_unix_nano);
```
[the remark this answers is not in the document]
lucastemb: Valid, but not worth fixing. Impact is narrow. If we were to miss a value under unique conditions (tight budgets after trimming), our goal to generate realistic data and loads still wouldn't be affected.

lading_payload/src/opentelemetry/metric.rs
```diff
@@ -109,14 +668,23 @@ pub struct MetricWeights {
...
         Self {
-            gauge: 50,          // 50%
-            sum_delta: 25,      // 25%
-            sum_cumulative: 25, // 25%
+            gauge: 40,                // 40%
+            sum_delta: 25,            // 20%
+            sum_cumulative: 25,       // 20%
+            histogram: 15,            //15%
+            exponential_histogram: 3, // 3%
+            summary: 2,               //2%
```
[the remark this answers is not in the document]
lucastemb: They should still be present regardless of whether or not a user opts out. The goal is ultimately to determine parity between ADP and the Agent. Users shouldn't be able to "opt-out" or "in" because that wasn't a feature that was present for counts and gauges before.

lading_payload/src/opentelemetry/metric.rs
```diff
@@ -403,29 +1016,274 @@ impl<'a> SizedGenerator<'a> for OpentelemetryMetrics {
...
+                                    state.clock.stream_start_time_unix_nano;
+                                point.time_unix_nano = current_time_unix_nano;
+                                state.cumulative_count += interval_population.count;
+                                state.cumulative_sum += interval_population.sum;
+
+                                let mut sorted_observations = interval_population.observations;
+                                sorted_observations.sort_by(f64::total_cmp);
+                                for quantile in &mut point.quantile_values {
+                                    quantile.value =
+                                        quantile_value(&sorted_observations, quantile.quantile);
```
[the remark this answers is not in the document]
lucastemb: This is intentional. OTLP summary `count` and `sum` are cumulative, but `quantile_values `describe the current snapshot. Here the latest interval is the snapshot, so computing quantiles from `interval_population.observations` is correct.

lading_payload/src/opentelemetry/metric/templates.rs
```diff
@@ -51,6 +62,67 @@ fn exponential_weighted_range<R: Rng + ?Sized>(rng: &mut R, min: u32, max: u32)
...
+    quantile_count: usize,
+    rng: &mut R,
+) -> Vec<f64> {
+    let mut quantiles: Vec<f64> = (0..quantile_count).map(|_| rng.random()).collect();
+    quantiles.sort_by(f64::total_cmp);
+    quantiles
+}
+
+pub(super) fn point_profile(seed: u64) -> PointProfile {
+    let sample = scaled_unit(seed, 0, 25.0, 1_000_000.0);
```
preinlein, line 75: Are these mins/maxes sourced from any spec or are they arbitrary?

I don't want to over-engineer (we don't necessarily need this to be configurable - if possible, the generator should do the "right" thing where possible) things here so if they're arbitrary, let's comment that.

lading_payload/src/opentelemetry/metric/templates.rs
```diff
@@ -51,6 +62,67 @@ fn exponential_weighted_range<R: Rng + ?Sized>(rng: &mut R, min: u32, max: u32)
...
+    let jitter = scaled_unit(seed, 42, 0.0, 10_000.0);
+
+    PointProfile {
+        sample,
+        spread,
+        jitter,
+    }
+}
+
+fn histogram_bounds_from_profile(profile: PointProfile, n_bounds: usize) -> Vec<f64> {
```
preinlein, line 86: imo this could be a function on PointProfile

I'd like to see some tests for this. There's enough magic numbers & special handling that I'd want this property tested to make sure we end up with the right values.

lading_payload/src/opentelemetry/metric/templates.rs
```diff
@@ -51,6 +62,67 @@ fn exponential_weighted_range<R: Rng + ?Sized>(rng: &mut R, min: u32, max: u32)
...
+
+    for idx in 0..n_bounds {
+        next += base_step + drift + (idx as f64 * 0.25);
+        bounds.push(next);
+    }
+
+    bounds
+}
+
+pub(super) fn zero_threshold_from_profile(profile: PointProfile) -> f64 {
```
preinlein, line 104: similar comment to above

(another reason I want tests is I could see us changing some of these magic numbers over time and I'd want a test in place already)

lading_payload/src/opentelemetry/metric/templates.rs
```diff
@@ -51,6 +62,67 @@ fn exponential_weighted_range<R: Rng + ?Sized>(rng: &mut R, min: u32, max: u32)
...
+    let candidate = (profile.spread / 512.0).max(profile.jitter / 1_024.0);
+    candidate.min(profile.spread / 4.0).max(0.0)
+}
+
+fn scaled_unit(seed: u64, shift: u32, min: f64, max: f64) -> f64 {
+    let value = ((seed.rotate_right(shift) & 0xFFFF) as f64) / 65_535.0;
```
preinlein, line 110: 65535 is the biggest value for a u16, why are we using that here exactly? I'm not following the logic

lading_payload/src/opentelemetry/metric/templates.rs
```diff
@@ -51,6 +62,67 @@ fn exponential_weighted_range<R: Rng + ?Sized>(rng: &mut R, min: u32, max: u32)
...
+    min + value * (max - min)
+}
+
+fn point_attributes(metadata: &[KeyValue], point_index: usize) -> Vec<KeyValue> {
+    let mut attributes = Vec::with_capacity(metadata.len().min(2) + 1);
```
preinlein, line 115: Why are we doing `metadata.len().min(2)` (the min(2) part) here?

aka why are we exclusively grabbing 2 key-value pairs and nothing else?

lading_payload/src/opentelemetry/metric/templates.rs
```diff
@@ -193,6 +277,101 @@ impl<'a> crate::SizedGenerator<'a> for MetricTemplateGenerator {
...
+                Data::Summary(metrics::v1::Summary { data_points })
+            }
+            Kind::Histogram {
+                aggregation_temporality,
+            } => {
+                let data_points = (0..total_data_points as usize)
+                    .map(|point_index| {
+                        let attributes = point_attributes(&metadata, point_index);
+                        let profile = point_profile(rng.random());
+                        let n_bounds = rng.random_range(4_usize..=16);
```
preinlein, line 312: What's this number determined from? Why are we using this specifically?

lading_payload/src/opentelemetry/metric/templates.rs
```diff
@@ -223,15 +402,59 @@ impl<'a> crate::SizedGenerator<'a> for MetricTemplateGenerator {
     }
 }
 
+#[cfg(test)]
+pub(super) fn random_partition<R: Rng + ?Sized>(
```
preinlein, line 406: what's this used for?

lading_payload/src/opentelemetry/metric.rs
```diff
@@ -60,8 +68,559 @@ use unit::UnitGenerator;
 /// `smallest_protobuf` test.
 pub const SMALLEST_PROTOBUF: usize = 31;
 
-/// Increment timestamps by 100 milliseconds (in nanoseconds) per tick
-const TIME_INCREMENT_NANOS: u64 = 1_000_000;
+/// Advance collection timestamps by 100 milliseconds per synthetic tick.
+const TIME_INCREMENT_NANOS: u64 = 100_000_000;
+const BASE_TIME_UNIX_NANOS: u64 = 1_700_000_000_000_000_000;
```
preinlein, line 73: Why this value?

lading_payload/src/opentelemetry/metric.rs
```diff
@@ -60,8 +68,559 @@ use unit::UnitGenerator;
...
+const DELTA_TEMPORALITY: i32 = 1;
```
preinlein, line 74: Are these values sourced from a spec?

What does 1/2 signify here?

lading_payload/src/opentelemetry/metric/templates.rs
```diff
@@ -23,6 +25,15 @@ use crate::{Error, Generator, common::config::ConfRange, common::strings};
 
 pub(crate) type Pool = templates::Pool<ResourceMetrics, ResourceTemplateGenerator>;
 
+const POINT_INDEX_ATTRIBUTE_KEY: &str = "lading.point_index";
+
+#[derive(Clone, Copy, Debug)]
+pub(super) struct PointProfile {
```
preinlein, line 31: Could we put a small description as to what this is used for?

What I'm picking up is that it describes the pool of points we can randomly pick from?

lading_payload/src/opentelemetry/metric.rs
```diff
@@ -60,8 +68,559 @@ use unit::UnitGenerator;
...
+        let bucket_index = explicit_bounds.partition_point(|bound| *bound < *observation);
+        if let Some(bucket_count) = bucket_counts.get_mut(bucket_index) {
+            *bucket_count += 1;
+        }
+    }
+    bucket_counts
+}
+
+fn merge_bucket_counts(target: &mut [u64], update: &[u64]) {
+    debug_assert_eq!(
```
preinlein, line 460: How do we imagine working with this?

imo I'd prefer to have property tests that exercise this in some way

lading_payload/src/opentelemetry/metric.rs
```diff
@@ -403,29 +1016,274 @@ impl<'a> SizedGenerator<'a> for OpentelemetryMetrics {
...
                     }
                 }
             }
         }
 
-        self.data_points_per_resource = data_points_count;
+        while tpl.encoded_len() > budget_before_fetch {
+            if !trim_one_data_point(&mut tpl) {
+                *budget = budget_before_fetch;
+                debug!(
```
preinlein, line 1259: Is this problematic or a "best effort" type of thing? Should these logs be error logs? 

I wouldn't want silent failures in prod if it's an actual problem of some kind.

lading_payload/src/opentelemetry/metric.rs
```diff
@@ -487,20 +1320,9 @@ impl crate::Serialize for OpentelemetryMetrics {
                     ?SMALLEST_PROTOBUF,
                     "to_bytes inner loop"
                 );
-                if required_bytes > max_bytes {
-                    total_data_points -= self.data_points_per_resource;
-                    drop(request.resource_metrics.pop());
-                    break;
-                }
+                debug_assert!(required_bytes <= max_bytes);
```
preinlein, line 1323: Do we need this to be a debug_assert? Could it be a real assertion?

I'm generally not a fan of debug assertions as I don't really know what to do with them when I encounter them: are they a real problem? Is it ok to silently fail/hit this code path?

Because if "it's fine for this to happen", why have the assertion at all.
If it's not fine, we should know about it.

lading_payload/src/opentelemetry/metric.rs
preinlein, line 1: I'd like to see more tests for the added functionality in here.

The internals of the opentelemetry metrics are not readily available for individuals working in lading so we need the tests to really help describe the expected behavior of things.

preinlein (commented):
I realize there's a lot of feedback on this PR, happy to pair or talk offline about em

PR #1858, opened 2026-04-14
New tcp based network workload

usamasaqib:
### What does this PR do?

This PR adds a new tcp based network workload in lading supporting an application level measure of throughput. The workload performs transactions (ping/pongs), and the number of transactions performed within an experiment duration is the throughput. This is tracked via responses received by the generator.

This is an almost 1-1 rewrite of the tcp_rr tests in [neper](https://github.com/google/neper), with some simplifications for our usecase.

This load generator offers few advantages on top of the already existing tcp load generator:
- User controlled number of threads and flows.
- A more efficient blackhole implementation which can handle multiple flows distributed over a set number of threads.
- Precisely tuning the networking stack to load the intended parts of the system, for example set `TCP_NODELAY` to avoid buffering data.

In the future more toggles can be exposed such as pinning threads, different algorithms for assigning flows to sockets on the server side, etc.

### Motivation

Support load testing eBPF based network tracers.

### DIFF
--- a/lading/src/bin/lading.rs
+++ b/lading/src/bin/lading.rs
@@ -543,11 +543,9 @@ async fn inner_main(
     //
     // GENERATOR
     //
-    let sample_period = Duration::from_millis(config.sample_period_milliseconds);
     for cfg in config.generator {
         let tgt_rcv = tgt_snd.subscribe();
-        let generator_server =
-            generator::Server::new(cfg, shutdown_watcher.clone(), sample_period)?;
+        let generator_server = generator::Server::new(cfg, shutdown_watcher.clone())?;
         gsrv_joinset.spawn(generator_server.run(tgt_rcv));
     }
 
@@ -566,8 +564,7 @@ async fn inner_main(
     // BLACKHOLE
     //
     for cfg in config.blackhole {
-        let blackhole_server =
-            blackhole::Server::new(cfg, shutdown_watcher.clone(), sample_period)?;
+        let blackhole_server = blackhole::Server::new(cfg, shutdown_watcher.clone())?;
         let _bsrv = tokio::spawn(async {
             match blackhole_server.run().await {
                 Ok(()) => debug!("blackhole shut down successfully"),
--- a/lading/src/bin/lading.rs
+++ b/lading/src/bin/lading.rs
@@ -543,9 +543,11 @@ async fn inner_main(
     //
     // GENERATOR
     //
+    let sample_period = Duration::from_millis(config.sample_period_milliseconds);
     for cfg in config.generator {
         let tgt_rcv = tgt_snd.subscribe();
-        let generator_server = generator::Server::new(cfg, shutdown_watcher.clone())?;
+        let generator_server =
+            generator::Server::new(cfg, shutdown_watcher.clone(), sample_period)?;
         gsrv_joinset.spawn(generator_server.run(tgt_rcv));
     }
 
@@ -564,7 +566,8 @@ async fn inner_main(
     // BLACKHOLE
     //
     for cfg in config.blackhole {
-        let blackhole_server = blackhole::Server::new(cfg, shutdown_watcher.clone())?;
+        let blackhole_server =
+            blackhole::Server::new(cfg, shutdown_watcher.clone(), sample_period)?;
         let _bsrv = tokio::spawn(async {
             match blackhole_server.run().await {
                 Ok(()) => debug!("blackhole shut down successfully"),
--- a/lading/src/blackhole/tcp_rr.rs
+++ b/lading/src/blackhole/tcp_rr.rs
@@ -15,6 +15,7 @@
 
 use std::io::{Read, Write};
 use std::net::SocketAddr;
+use std::num::{NonZeroU16, NonZeroUsize};
 use std::os::fd::{AsRawFd, FromRawFd};
 use std::sync::atomic::Ordering::Relaxed;
 use std::sync::{Arc, Barrier};
@@ -31,12 +32,14 @@ use crate::neper::flow::{self, Action, Flow, FlowMap};
 use crate::neper::metrics::{self, ThreadMetrics};
 use crate::neper::thread;
 
-fn default_one() -> u16 {
-    1
+const fn default_nonzero_u16() -> NonZeroU16 {
+    // Safety: 1 != 0
+    unsafe { NonZeroU16::new_unchecked(1) }
 }
 
-fn default_one_usize() -> usize {
-    1
+const fn default_nonzero_usize() -> NonZeroUsize {
+    // Safety: 1 != 0
+    unsafe { NonZeroUsize::new_unchecked(1) }
 }
 
 fn default_control_port() -> u16 {
@@ -58,14 +61,14 @@ pub struct Config {
     pub control_port: u16,
     /// Number of OS server threads. Default 1. When > 1, uses `SO_REUSEPORT`
     /// with an eBPF program for load balancing
-    #[serde(default = "default_one")]
-    pub threads: u16,
+    #[serde(default = "default_nonzero_u16")]
+    pub threads: NonZeroU16,
     /// Bytes to read per request. Default 1.
-    #[serde(default = "default_one_usize")]
-    pub request_size: usize,
+    #[serde(default = "default_nonzero_usize")]
+    pub request_size: NonZeroUsize,
     /// Bytes to send per response. Default 1.
-    #[serde(default = "default_one_usize")]
-    pub response_size: usize,
+    #[serde(default = "default_nonzero_usize")]
+    pub response_size: NonZeroUsize,
     /// Whether to set `TCP_NODELAY` on accepted connections. Default true.
     #[serde(default = "default_true")]
     pub no_delay: bool,
@@ -144,7 +147,7 @@ impl TcpRr {
     /// Panics if the ready-barrier tokio task is cancelled.
     pub async fn run(self) -> Result<(), Error> {
         let shutdown_flag = thread::new_shutdown_flag();
-        let num_threads = self.config.threads;
+        let num_threads = self.config.threads.get();
 
         let thread_metrics = Arc::new(
             (0..num_threads)
@@ -178,8 +181,8 @@ impl TcpRr {
         let mut handles = Vec::with_capacity(num_threads as usize);
         for i in 0..num_threads {
             let binding_addr = self.config.binding_addr;
-            let request_size = self.config.request_size;
-            let response_size = self.config.response_size;
+            let request_size = self.config.request_size.get();
+            let response_size = self.config.response_size.get();
             let no_delay = self.config.no_delay;
             let flag = Arc::clone(&shutdown_flag);
             let tm = Arc::clone(&thread_metrics);
--- a/lading/src/blackhole/tcp_rr.rs
+++ b/lading/src/blackhole/tcp_rr.rs
@@ -17,8 +17,8 @@ use std::io::{Read, Write};
 use std::net::SocketAddr;
 use std::num::{NonZeroU16, NonZeroUsize};
 use std::os::fd::{AsRawFd, FromRawFd};
+use std::sync::Arc;
 use std::sync::atomic::Ordering::Relaxed;
-use std::sync::{Arc, Barrier};
 use std::time::Duration;
 
 use mio::net::TcpStream;
@@ -163,10 +163,13 @@ impl TcpRr {
             })
         };
 
-        // BPF barrier: thread 0 must bind first to attach BPF before other
-        // threads join the SO_REUSEPORT group.
-        let bpf_barrier = if num_threads > 1 {
-            Some(Arc::new(Barrier::new(num_threads as usize)))
+        // Pre-build thread 0's listener here so the BPF program is attached
+        // to the reuseport group before any other thread calls bind(). This
+        // removes the need for a cross-thread BPF barrier — if the bind fails
+        // or panics, it propagates as an error directly from this task.
+        let binding_addr = SocketAddr::new(self.config.addr, self.config.data_port);
+        let thread0_listener = if num_threads > 1 {
+            Some(create_listener(0, num_threads, binding_addr))
         } else {
             None
         };
@@ -178,26 +181,26 @@ impl TcpRr {
         let (ready_tx, mut ready_rx) = tokio::sync::mpsc::unbounded_channel::<()>();
 
         let mut handles = Vec::with_capacity(num_threads as usize);
+        let mut thread0_listener = thread0_listener;
         for i in 0..num_threads {
-            let binding_addr = SocketAddr::new(self.config.addr, self.config.data_port);
             let request_size = self.config.request_size.get();
             let response_size = self.config.response_size.get();
             let no_delay = self.config.no_delay;
             let flag = Arc::clone(&shutdown_flag);
             let tm = Arc::clone(&thread_metrics);
-            let bpf_barrier = bpf_barrier.clone();
+            let prebuilt = if i == 0 { thread0_listener.take() } else { None };
             let tx = ready_tx.clone();
             let handle = thread::spawn_named(&format!("tcp_rr-server-{i}"), move || {
                 server_thread_main(
                     i,
                     num_threads,
                     binding_addr,
+                    prebuilt,
                     request_size,
                     response_size,
                     no_delay,
                     &flag,
                     &tm[i as usize],
-                    bpf_barrier.as_deref(),
                     tx,
                 );
             });
@@ -382,30 +385,18 @@ fn server_thread_main(
     thread_index: u16,
     num_threads: u16,
     binding_addr: SocketAddr,
+    prebuilt_listener: Option<std::net::TcpListener>,
     request_size: usize,
     response_size: usize,
     no_delay: bool,
     shutdown_flag: &std::sync::atomic::AtomicBool,
     metrics: &ThreadMetrics,
-    bpf_barrier: Option<&Barrier>,
     ready_tx: tokio::sync::mpsc::UnboundedSender<()>,
 ) {
-    // Thread 0 must bind first (to attach BPF before others join the
-    // reuseport group). Non-zero threads wait at the BPF barrier first.
-    if let Some(barrier) = bpf_barrier
-        && thread_index != 0
-    {
-        barrier.wait();
-    }
-
-    let std_listener = create_listener(thread_index, num_threads, binding_addr);
-
-    // Thread 0 signals the BPF barrier after binding so others can proceed.
-    if let Some(barrier) = bpf_barrier
-        && thread_index == 0
-    {
-        barrier.wait();
-    }
+    // Thread 0 uses the pre-built listener (with BPF already attached);
+    // others bind their own sockets that join the existing reuseport group.
+    let std_listener = prebuilt_listener
+        .unwrap_or_else(|| create_listener(thread_index, num_threads, binding_addr));
 
     // Signal that this thread's listener is bound and ready. If this send
     // fails the receiver has gone away (blackhole is shutting down).
--- a/lading/src/blackhole/tcp_rr.rs
+++ b/lading/src/blackhole/tcp_rr.rs
@@ -16,7 +16,7 @@
 use std::io::{Read, Write};
 use std::net::SocketAddr;
 use std::num::{NonZeroU16, NonZeroUsize};
-use std::os::fd::{AsRawFd, FromRawFd};
+use std::os::fd::AsRawFd;
 use std::sync::Arc;
 use std::sync::atomic::Ordering::Relaxed;
 use std::time::Duration;
@@ -50,6 +50,10 @@ fn default_data_port() -> u16 {
     12867
 }
 
+fn default_backlog() -> i32 {
+    1024
+}
+
 const fn default_true() -> bool {
     true
 }
@@ -79,6 +83,10 @@ pub struct Config {
     /// Whether to set `TCP_NODELAY` on accepted connections. Default true.
     #[serde(default = "default_true")]
     pub no_delay: bool,
+    /// Listener backlog (pending-connection queue length) passed to `listen(2)`.
+    /// Default 1024.
+    #[serde(default = "default_backlog")]
+    pub backlog: i32,
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -169,7 +177,7 @@ impl TcpRr {
         // or panics, it propagates as an error directly from this task.
         let binding_addr = SocketAddr::new(self.config.addr, self.config.data_port);
         let thread0_listener = if num_threads > 1 {
-            Some(create_listener(0, num_threads, binding_addr))
+            Some(create_listener(0, num_threads, binding_addr, self.config.backlog))
         } else {
             None
         };
@@ -186,6 +194,7 @@ impl TcpRr {
             let request_size = self.config.request_size.get();
             let response_size = self.config.response_size.get();
             let no_delay = self.config.no_delay;
+            let backlog = self.config.backlog;
             let flag = Arc::clone(&shutdown_flag);
             let tm = Arc::clone(&thread_metrics);
             let prebuilt = if i == 0 {
@@ -200,6 +209,7 @@ impl TcpRr {
                     num_threads,
                     binding_addr,
                     prebuilt,
+                    backlog,
                     request_size,
                     response_size,
                     no_delay,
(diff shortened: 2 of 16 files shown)


lading/src/neper/metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub(crate) const fn new() -> Self {
+        Self {
+            value: UnsafeCell::new(0),
+        }
+    }
+
+    /// Increment. Must only be called from the owning thread.
+    #[inline]
+    pub(crate) fn add(&self, n: u64) {
+        unsafe { *self.value.get() += n }
```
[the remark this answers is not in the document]
usamasaqib: In local experiments using atomics showed measurable performance overhead, in the form of lower response_received/sec. I will let someone else weigh in here about the real risk.

lading/src/generator/tcp_rr.rs
```diff
@@ -0,0 +1,293 @@
...
+) {
+    let mut poll = Poll::new().expect("failed to create mio::Poll");
+    let mut events = Events::with_capacity(num_flows as usize);
+    let request_buf = vec![0u8; request_size];
+    let mut response_buf = vec![0u8; response_size];
+    let mut flows: FlowMap<ClientState> = FlowMap::new();
+    let mut next_token: usize = 0;
+
+    for _ in 0..num_flows {
+        match std::net::TcpStream::connect(addr) {
```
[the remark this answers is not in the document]
usamasaqib: Not a concern. It is fine if this happens and we block for a while. We will surface the errors eventually and the problem can be addressed.

lading/src/bin/lading.rs
```diff
@@ -543,9 +543,11 @@ async fn inner_main(
     //
     // GENERATOR
     //
+    let sample_period = Duration::from_millis(config.sample_period_milliseconds);
```
goxberry: What's the motivation for changing the sample period from 1 s? The assumption of a sampling period of 1 s is tightly integrated in a number of places throughout Lading and its consumers, so I view this change as high risk.

If we were to expose this feature from `tcp_rr` blackholes and generators, I think we would want `sample_period` to be a field of `lading::generator::tcp_rr::Config` and a field of `lading::blackhole::tcp_rr::Config` to preserve the existing `generator::Server::new` and `blackhole::Server::new` interfaces.

lading/src/neper/metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    let mut total_snapshot_ns: u64 = 0;
+    let mut snapshot_count: u64 = 0;
+
+    while !shutdown.load(Relaxed) {
+        std::thread::sleep(sample_period);
+        let start = Instant::now();
+        snapshot.snapshot_and_submit(thread_metrics, labels);
+        total_snapshot_ns += u64::from(start.elapsed().subsec_nanos());
+        snapshot_count += 1;
```
goxberry: I think as written, this function will register counters roughly `sample_period` milliseconds apart (plus whatever overhead is incurred in `snaphsot.snapshot_and_submit`). Captures written to file from the metrics recorder's registry are still [written at an interval of approximately 1,000 ms](https://github.com/DataDog/lading/blob/5d23e9166edecbad14e5ac8461d1cf477924253b/lading/src/target.rs#L334-L347), which is the [metrics flush interval](https://github.com/DataDog/lading/blob/5d23e9166edecbad14e5ac8461d1cf477924253b/lading_capture/src/manager.rs#L34-L36).

Is this behavior what you intend?

lading/src/bin/lading.rs
usamasaqib: The motivation is reduce the impact of the snapshotter thread on the system. The snapshotter will do more work as  more flows are added, and at some point it will essentially amount to busy looping of all flows as the read time approaches 1 second.

I am okay with your suggestion of making the snapshot interval specific to this workload's configuration.

lading/src/blackhole/tcp_rr.rs
```diff
@@ -0,0 +1,446 @@
...
+    request_buf: &mut [u8],
+    response_buf: &[u8],
+    metrics: &ThreadMetrics,
+) -> Action {
+    match flow.state {
+        ServerState::RecvRequest => {
+            let offset = request_buf.len() - flow.xfer;
+            match flow.stream.read(&mut request_buf[offset..]) {
+                Ok(0) => Action::Remove,
+                Ok(n) => {
```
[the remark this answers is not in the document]
usamasaqib: zero value for request_size is rejected now

lading/src/generator/tcp_rr.rs
```diff
@@ -0,0 +1,293 @@
...
+                    stream,
+                    token,
+                    state: ClientState::SendRequest,
+                    xfer: request_size,
+                });
+            }
+            Err(e) => {
+                trace!("connection to {addr} failed: {e}");
+                metrics.connections_failed.add(1);
+            }
```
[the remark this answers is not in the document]
usamasaqib: Using a control port to synchronize startup

lading/src/generator/tcp_rr.rs
```diff
@@ -0,0 +1,293 @@
...
+                Err(e) => {
+                    trace!("write error: {e}");
+                    Action::Remove
+                }
+            }
+        }
+        ClientState::RecvResponse => {
+            let offset = response_buf.len() - flow.xfer;
+            match flow.stream.read(&mut response_buf[offset..]) {
+                Ok(0) => Action::Remove,
```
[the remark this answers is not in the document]
usamasaqib: zero for response_size rejected now

lading/src/blackhole/tcp_rr.rs
```diff
@@ -0,0 +1,442 @@
...
+        // ordering: thread 0 binds + attaches, then signals the barrier,
+        // then all other threads proceed to bind.
+        let barrier = if num_threads > 1 {
+            Some(Arc::new(Barrier::new(num_threads as usize)))
+        } else {
+            None
+        };
+
+        let mut handles = Vec::with_capacity(num_threads as usize);
+        for i in 0..num_threads {
```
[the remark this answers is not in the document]
usamasaqib: zero for threads rejected now

lading/src/blackhole/tcp_rr.rs
```diff
@@ -0,0 +1,499 @@
...
+        // Wait for all data threads to finish binding.
+        // Use spawn_blocking so we don't block the tokio runtime.
+        let rb = Arc::clone(&ready_barrier);
+        tokio::task::spawn_blocking(move || rb.wait()).await.expect("ready barrier join failed");
+
+        // All data listeners are up. Open control port so the generator
+        // can connect and know we're ready.
+        let control_addr = SocketAddr::new(self.config.addr, self.config.control_port);
+        let control_listener = std::net::TcpListener::bind(control_addr)
+            .map_err(|source| Error::Bind { addr: control_addr, source: Box::new(source) })?;
```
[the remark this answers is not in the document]
usamasaqib: That isn't a problem. We do not need this level of resiliency.

lading/src/neper/metrics.rs
usamasaqib: The goal was to reduce the overhead associated with the snapshotter thread. Since we do not need per second granularity I made this configurable.

What will the effect of this be on the data itself?

goxberry:
From a review standpoint, I suspect this PR would benefit from splitting it into a PR on the neper-like generator/blackhole pair and a PR on the sampling period because the sampling period changes have the potential to be cross-cutting and might require additional care, depending on the intended outcome.

For the neper-like components, I'd like to understand more the motivation for a new generator and blackhole, and the motivation for the particular concurrency model in the implementation. Some of what would help me understand the motivation would be things like:

- you and/or your team have a lot of experience working with neper-like tools
- what you can do with the generator/blackhole pair here that couldn't be done with, say, the existing TCP generator/blackhole pair
- the distinction between an application-level measure of throughput versus the data we collect now

I'm also interested in understanding the motivation for using a `std::thread` + `std::sync` + `mio` concurrency model inside the generator and inside the blackhole, as opposed to, say, `tokio`. I'd particularly like to understand how the benefits of using a second concurrency model stack up against the risks of its complexity. Are there benchmarking cases where the results indicate a `tokio`-based approach would be insufficient? From a maintenance perspective, it's challenging for me to reason about this composition of the `tokio` model used by the rest of Lading and the concurrency model used within the new generator/blackhole pair. For instance, the new generator, if configured, would be spawned as a `tokio` task with `tokio::join_set::JoinSet::spawn` in the generator server start-up loop. There appear to be few `await` points in the new generator, and it looks like the only channel-like value being used is the shutdown watcher. I'm curious as to whether the new generator does enough computation to block the `tokio` scheduler from making progress.

lading/src/neper/metrics.rs
goxberry: If the `sample_period` value is at least 1,000, I'd expect data collection to proceed roughly as it does now. If it is less than 1,000, then I'd expect to see counters aggregate because our capture manager is hard coded to write metrics captures approximately one second apart.

usamasaqib: So it should be enough to set the lower bound of the sample_period to 1000?

usamasaqib:
> From a review standpoint, I suspect this PR would benefit from splitting it into a PR on the neper-like generator/blackhole pair and a PR on the sampling period because the sampling period changes have the potential to be cross-cutting and might require additional care, depending on the intended outcome.
> 

Sounds good.

> For the neper-like components, I'd like to understand more the motivation for a new generator and blackhole, and the motivation for the particular concurrency model in the implementation. Some of what would help me understand the motivation would be things like:
> 
> * you and/or your team have a lot of experience working with neper-like tools
> * what you can do with the generator/blackhole pair here that couldn't be done with, say, the existing TCP generator/blackhole pair

Primarily this new setup allows more control over the shape of the network traffic. As a first step we get to have the ability to specify exactly how many flows we want and how many threads we want them distributed over. With regards to the blackhole, unlike the current implementation, we also get the ability to handle multiple flows in parallel by the use of `SO_REUSEPORT` option, which allows for certain tests which cannot be done with the current setup.

In the future we want to be able to control more details of the traffic, like some of the examples below.

> * the distinction between an application-level measure of throughput versus the data we collect now
> 

Currently what is collected, in the tcp workload, is the amount of bytes written/read to/from the socket. This lacks a lot of details that we need in order to be able to make claims about specific code paths in the networking stack, and potentially suffers from being too noisy in a multi-threaded environment.
bytes_written does not let us determine how many packets were actually pushed out, how much of the data was buffered before being pushed out, where it was queued, and where it gets batched, etc. All of this information is essential for us to be able to predict the load we are pushing on specific ebpf programs.
The current implementation allows us to control for these by setting various socket options, and controlling the amount of data written per transaction.

The application level measure of throughput comes from the number of transaction we are able to complete in a set amount of time. A transaction is a request sent and a response received. The rate of transactions allows us to determine the overhead we are adding in a specific code path in the networking stack. New experiments in the future will use this infrastructure to exercise different parts of the stack we are interested in.

> I'm also interested in understanding the motivation for using a `std::thread` + `std::sync` + `mio` concurrency model inside the generator and inside the blackhole, as opposed to, say, `tokio`. I'd particularly like to understand how the benefits of using a second concurrency model stack up against the risks of its complexity. Are there benchmarking cases where the results indicate a `tokio`-based approach would be insufficient? 

We have several reasons for relying on threads:
1. tokio gives us the abstraction of tasks, which may not map 1:1 to os-threads. We want to know exactly how many os-threads we are launching to control parallelism and ensure repeatability.
2. We want to control exactly the placement of these os-threads for certain experiments. For example, we want to measure workload impact in cases where packet reception is happening on the set of CPUs separate from the ones where NIC Rx queues are mapped.
3. Os threads can be easily assigned to cgroups for even more specific setups.

> From a maintenance perspective, it's challenging for me to reason about this composition of the `tokio` model used by the rest of Lading and the concurrency model used within the new generator/blackhole pair. For instance, the new generator, if configured, would be spawned as a `tokio` task with `tokio::join_set::JoinSet::spawn` in the generator server start-up loop. There appear to be few `await` points in the new generator, and it looks like the only channel-like value being used is the shutdown watcher. I'm curious as to whether the new generator does enough computation to block the `tokio` scheduler from making progress.

There are a few blocking things in the async function right now. If you have concerns about these we can make them async as well. Other than that all blocking operations are done within threads outside of the tokio runtime.

lading/src/neper/metrics.rs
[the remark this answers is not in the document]
goxberry: To me, at least, if we're to rely on those metrics, I'd favor correctness or data integrity over performance unless there's an extremely compelling argument to the contrary. We have limited capabilities to detect "bad" data -- mostly due to test rig misconfigurations, and those capabilities aren't all automated because I haven't thought about how to automate some of those checks.

My guess is that if data races were to happen and these only affected the collected metrics, we would misattribute it to "noise in the system", so it would get lumped in with stuff we can't control as Gaussian randomness. That's not necessarily a dealbreaker to me so much as it would be hard to identify this contribution after the fact, and "just part of the Gaussian noise" may be mis-specification that contributes to false results (positive or negative). If this implementation is kept, it would be worth noting these data race issues in the comments.

lading/src/blackhole/tcp_rr.rs
```diff
@@ -0,0 +1,534 @@
...
+            });
+            handles.push(handle);
+        }
+
+        // Wait for all data threads to finish binding.
+        // Use spawn_blocking so we don't block the tokio runtime.
+        let rb = Arc::clone(&ready_barrier);
+        tokio::task::spawn_blocking(move || rb.wait())
+            .await
+            .expect("ready barrier join failed");
```
goxberry: If one of the `server_thread_main` threads panic before reaching `ready_barrier.wait()` inside the `server_thread_main` call, I believe the `ready_barrier.wait()` call hangs indefinitely, and this task spawned in the `tokio` runtime's blocking thread pool will run indefinitely. In that scenario, the task spawned by `spawn_blocking` also can't be aborted, which might complicate the `tokio` runtime shutdown process.

As an alternative, you might consider having each of the `server_thread_main` threads send a signal over a channel to the `tokio` task (or some other one-way mechanism) signaling that each thread is either "ready" or "in an error condition on thread `thread_index`", and have the `tokio` task wait until `num_threads` messages are received.

lading/src/blackhole/tcp_rr.rs
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    let std_listener = create_listener(thread_index, num_threads, binding_addr);
+
+    // Thread 0 signals the BPF barrier after binding so others can proceed.
+    if let Some(barrier) = bpf_barrier
+        && thread_index == 0
+    {
+        barrier.wait();
+    }
```
goxberry: As with https://github.com/DataDog/lading/pull/1858/changes#r3090672847, I think there's a similar concern about `bpf_barrier` deadlocking if `create_listener` panics when `thread_index == 0`. An alternative might be something like a `std::sync::Mutex` to store the success/failure condition of the `create_listener` operation plus a `std::sync::CondVar` to notify all threads when thread index zero modifies the value stored in the `Mutex`.

goxberry:
Okay, that explanation of the need to place the workload on OS threads is compelling. From a concurrency perspective, I think the only other issues I see involve potential fault-tolerance issues with the `std::sync::Barrier`s used in the blackhole, so I've left comments on those along with possible alternatives.

Concurrency is not an area I'm expert at, so I'll ask for a second set of eyes on the barriers that I flagged, just to confirm.

lading/src/neper/metrics.rs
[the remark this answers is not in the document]
goxberry: Now that I think about it, we're also starting to use Lading in correctness testing rigs, too, and for that use case, corrupted counters would be problematic.

Could you give me a sense of how meaningful the overhead would be for the applications you're thinking about?

lading/src/bin/lading.rs
preinlein: Question, are you thinking that folks will determine this themselves in the config?

My own 2 cents is that I would like lading to be as smart as possible and for it to configure itself accordingly or crash early if the config is unreasonable. It's something I've spoken to @blt about: we make claims that lading is faster than the target but we don't necessarily enforce it.

We have a lot of unbounded config that can set lading in a mode that it would undoubtedly be slower than certain targets.

Although we haven't really tackled this problem yet, I would be hesitant to add a config that users need to be aware of that controls the internals of lading in this way. 

I do see where you're coming from and it is a valid concern, I'm just wondering if we don't want to tackle it holistically across lading as it's own thing.

lading/src/neper/metrics.rs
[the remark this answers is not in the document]
usamasaqib: The data race is benign though, so correctness should not be a concern?

I will get you some numbers soon

lading/src/blackhole/tcp_rr.rs
usamasaqib: Make sense, I will try out your suggestion.

lading/src/blackhole/tcp_rr.rs
```diff
@@ -0,0 +1,535 @@
+//! TCP request/response (`tcp_rr`) blackhole — the server side.
+//! Based on <https://github.com/google/neper>
```
preinlein, line 2: great comment, love breadcrumb comments

lading/src/blackhole/tcp_rr.rs
```diff
@@ -0,0 +1,535 @@
...
+    num_threads: u16,
+    binding_addr: SocketAddr,
+) -> std::net::TcpListener {
+    // Create socket manually to set SO_REUSEPORT before bind.
+    let domain = if binding_addr.is_ipv4() {
+        libc::AF_INET
+    } else {
+        libc::AF_INET6
+    };
+    let fd = unsafe { libc::socket(domain, libc::SOCK_STREAM, 0) };
```
GeorgeHahn: Would you mind switching these calls over to the `socket2` crate? It wraps these into a nicer API to work with.

lading/src/generator/tcp_rr.rs
```diff
@@ -0,0 +1,348 @@
...
+
+use super::General;
+use crate::generator::common::MetricsBuilder;
+use crate::neper::flow::{self, Action, Flow, FlowMap};
+use crate::neper::metrics::{self, ThreadMetrics};
+use crate::neper::thread;
+
+const fn default_nonzero_u16() -> NonZeroU16 {
+    // Safety: 1 != 0
+    unsafe { NonZeroU16::new_unchecked(1) }
```
GeorgeHahn: I'd prefer `NonZeroU16::new(1).expect(..)` - these are clearly correct by construction, but it's one less `unsafe` to audit.

lading/src/neper/bpf.rs
```diff
@@ -0,0 +1,145 @@
+//! eBPF program for `SO_REUSEPORT` load balancing.
+//!
+//! Provides an eBPF program matching neper's `prog1`: select random
+//! socket
```
GeorgeHahn, line 4: This is interesting and I'd like to understand it better. Could you expand this doc comment with context around where one would use this and why?

lading/src/neper/bpf.rs
```diff
@@ -0,0 +1,145 @@
...
+        libc::syscall(
+            libc::SYS_bpf,
+            BPF_PROG_LOAD,
+            (&raw const attr).cast::<libc::c_void>(),
+            std::mem::size_of::<BpfAttrProgLoad>() as libc::c_int,
+        )
+    };
+
+    if fd < 0 {
+        return Err(io::Error::last_os_error());
```
GeorgeHahn: Does this return decent errors? Should we wrap this up in an error of our own that can directly translate to 'bpf load failed: {reason}'?

lading/src/blackhole/tcp_rr.rs
```diff
@@ -0,0 +1,535 @@
...
+use std::io::{Read, Write};
+use std::net::SocketAddr;
+use std::num::{NonZeroU16, NonZeroUsize};
+use std::os::fd::{AsRawFd, FromRawFd};
+use std::sync::atomic::Ordering::Relaxed;
+use std::sync::{Arc, Barrier};
+use std::time::Duration;
+
+use mio::net::TcpStream;
+use mio::{Events, Interest, Poll, Token};
```
blt: I was surprised to see mio present in this module, that's unusual for this project. Could you leave a comment explaining why we can't use tokio only here?

lading/src/neper/metrics.rs
goxberry: Yes, I believe so.

lading/src/blackhole/tcp_rr.rs
usamasaqib: The primary reason is that we need control over which os-threads work is scheduled on, and tokio abstracts away os-threads.
I have more detail in this comment: https://github.com/DataDog/lading/pull/1858#issuecomment-4253323819

Let me know if anything is'nt clear, or if my understanding is incorrect.

lading/src/neper/bpf.rs
usamasaqib: Getting the `reason` can be a bit involved, and the bpf program at this stage is so simple that I do not think we need to add that. I can add wrap it up into something nicer to read though.

lading/src/neper/bpf.rs
usamasaqib: `SO_REUSEPORT` allows multiple sockets to be listening on the same port. This means that we can more effectively load a test server by loading it with multiple clients, and have multiple worker threads each with their own receive queue. Whenever a packet is received the kernel assigns it a socket (so effectively a worker thread). 
The bpf program here controls the assignment algorithm. By default the kernel hashes the connection tuple (src, dst, src_port, dst_port) and uses that to select a socket. We do not necessarily want to depend on this for two reasons
1. It is unclear what the distribution of the kernel function, jhash, is like and whether the load is evenly distributed.
2. Repeatability might suffer since the client ports are randomly assigned and that may change the hash functions distribution.

Today, the bpf program just randomly select a socket, since it is just intended as a placeholder for now. In the future we may use a different algorithm for selecting sockets, or expose that to users to pick an algorithm they think is appropriate for their workload.

lading/src/blackhole/tcp_rr.rs
usamasaqib: For some reason I can't respond to the above comment, so I will answer here.

The reason for adding the configuration here is to reduce the overhead added by the snapshotter thread when we have a large number of worker threads. @goxberry has some concerns about this as well. I am going to give it more thought and see if this can be reworked.

lading/src/neper/metrics.rs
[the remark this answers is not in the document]
usamasaqib: I have updated the code to use atomics now. Local benchmarks did not show meaningful difference, especially when setting the sampling duration to be higher

lading/src/neper/metrics.rs
usamasaqib: I am no longer passing sample duration to the generators and blackholes

lading/src/blackhole/tcp_rr.rs
usamasaqib: I opted to create the listener for thread 0 before launching any threads. This avoid the need to do any explicit synchronization.
https://github.com/DataDog/lading/pull/1858/changes/656bad6c823788950e5c4cded62aaefc3285f6e7

lading/src/blackhole/tcp_rr.rs
usamasaqib: Done. https://github.com/DataDog/lading/pull/1858/commits/392ad2f7471bf1605a185c6a1a5f15af0c1e7ad2

lading/src/neper/bpf.rs
usamasaqib: I have made the changes to extract the verifier log on load failure and print that out in the error as the reason.
https://github.com/DataDog/lading/pull/1858/commits/fd6ed597ed979d690164c934bad22c48f44dfaea

lading/src/blackhole/tcp_rr.rs
usamasaqib: I have removed the configurable snapshot duration. I think this can be revisited in the future when we better understand the usage of this new workload.

lading/src/neper/metrics.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+define_thread_metrics! {
+    requests_sent,
+    responses_received,
+    bytes_written,
+    bytes_read,
+    connections_failed,
+    connections_accepted,
+    requests_received_count,
```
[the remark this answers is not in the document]
goxberry: @usamasaqib Is this Codex comment accurate?

goxberry (approved):
LGTM; left a minor comment about a metric name.

lading/src/neper/metrics.rs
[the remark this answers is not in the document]
usamasaqib: It is not correct in that `tcp_rr` uses `requests_received`. It uses `reqeuets_received_count`. And the point about dashboards breaking is not a concern either because we have seperate widgets for ebpf experiments.

However, for naming consistency I can change it to `requests_received`.

PR #1868, opened 2026-05-06
Add configurable unique_tag_ratio to OTel metric Contexts

ArunPiduguDD:
## Summary

- Adds `unique_tag_ratio: f32` field to the `Contexts` struct in the OpenTelemetry metrics payload generator
- Replaces the hardcoded `UNIQUE_TAG_RATIO = 0.75` constant in `templates.rs` with the user-supplied value
- Validates the field is in the range `0.01–1.0`
- Defaults to `0.75` (preserving existing behavior)

## Usage

```yaml
contexts:
  unique_tag_ratio: 1.0  # every attribute is unique per emission
```

## Test plan

- [ ] All existing tests pass (`cargo test -p lading-payload`)
- [ ] Build clean (`cargo build -p lading-payload`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 - Updated to rand 0.10.x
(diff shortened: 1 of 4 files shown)


ArunPiduguDD:
Closing in favor of #1869. Per follow-up discussion, the unique_tag_ratio knob isn't needed for the cardinality use case — 0.75 default is sufficient. The remaining cardinality bottleneck is the u8 cap, addressed in #1869.

PR #1869, opened 2026-05-07
Widen scopes_per_resource and metrics_per_scope to u16 in OTel metrics config

ArunPiduguDD:
## Summary

- Widens `scopes_per_resource` and `metrics_per_scope` in `lading_payload::opentelemetry::metric::Contexts` from `ConfRange<u8>` to `ConfRange<u16>`
- Updates validation arithmetic to use `saturating_mul` so `u16 * u16` doesn't overflow `u32`
- Updates proptest ranges to use `u16` types

## Why

The previous `u8` typing capped per-emission cardinality at 255 × 255 = **65,025 unique metric sites per resource**. With this change, that ceiling rises to ~4.3 billion, which fits within the existing `total_contexts: u32` envelope. This unblocks high-cardinality OTel load tests that need many distinct metric time series per emission.

The `attributes_per_*` fields are intentionally left as `u8` — 255 tags per metric/resource/scope is plenty, and widening them would cascade into the shared `TagGenerator` (which OTel logs also depends on).

## Test plan

- [ ] All existing OTel metric tests pass (`cargo test -p lading-payload --lib opentelemetry::metric`)
- [ ] Workspace builds clean (`cargo build --workspace`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 - Updated to rand 0.10.x
+- Widened `scopes_per_resource` and `metrics_per_scope` in the OpenTelemetry metrics `Contexts` config from `u8` to `u16`, raising the per-emission cardinality ceiling from 65,025 to ~4.3B unique metric sites.
 
 ## [0.32.0]
 ## Changed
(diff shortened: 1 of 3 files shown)


ArunPiduguDD:
Closing — on reflection, the practical cardinality ceiling is memory-bound at ~10M unique time series, which was already reachable pre-PR by combining multiple resources (boxes) with per-resource items. The type widening doesn't raise the reachable ceiling, only the ergonomic flexibility in shaping that cardinality.

PR #1871, opened 2026-05-07
feat(dogstatsd): add configurable pools for origin detection extension fields

thieman:
## Summary

- Adds three new optional `Config` fields to the DogStatsD payload generator, each holding a pool of possible values for one of the origin detection protocol extension fields
- Per generated metric, a value is randomly selected from the pool or the field is omitted entirely — consistent with the existing behavior of `|c:` with its internal random pool
- Because millstone's `Payload::DogStatsD` wraps `lading_payload::dogstatsd::Config` directly via serde, all three fields are immediately usable in millstone's YAML configuration with no changes to millstone required
- No breaking changes: all existing `Config` constructions use `..Default::default()` or `Config::default()`; the metric structs cannot be externally constructed due to `pub(crate)` fields; internal constructors are all crate-private
- Fixes `choose_or_not_ref` to short-circuit before touching the RNG when the pool is empty, preserving determinism for seeded output when optional fields are unconfigured

### New fields

**`container_ids`** — pool for the `|c:` Local Data field. When empty, falls back to the prior behavior of using an internally generated random value. When non-empty, only values from the pool are used, so callers can supply real container IDs.

**`external_data`** — pool for the `|e:` External Data field. Empty by default; field is never emitted when empty.

**`cardinality`** — pool for the `|card:` cardinality override field. Empty by default; field is never emitted when empty.

### Fuzz corpus note

`Config` derives `arbitrary::Arbitrary` and is used as a fuzz input in two targets. Adding fields changes what the fuzzer generates, so saved fuzz corpora for those targets should be reset.

## Test plan

- [x] All existing `lading-payload` tests pass
- [x] New tests verify fields are emitted when pools are non-empty, absent when empty, and that emitted values come only from the configured pool
- [x] Probabilistic emission tests run 100 seeds (p(false negative) ≈ 10^-30)

### DIFF
--- a/lading_payload/src/common/strings.rs
+++ b/lading_payload/src/common/strings.rs
@@ -62,6 +62,9 @@ pub(crate) fn choose_or_not_ref<'a, R, T>(mut rng: &mut R, pool: &'a [T]) -> Opt
 where
     R: rand::Rng + ?Sized,
 {
+    if pool.is_empty() {
+        return None;
+    }
     if rng.random() {
         pool.choose(&mut rng)
     } else {
--- a/lading_payload/src/dogstatsd.rs
+++ b/lading_payload/src/dogstatsd.rs
@@ -198,6 +198,23 @@ pub struct Config {
     pub tag_names: Vec<String>,
     /// A list of possible tag values to generate
     pub tag_values: Vec<String>,
+    /// Pool of possible container ID values for the `|c:` origin detection extension field.
+    ///
+    /// When non-empty, each generated metric randomly selects from this pool (or omits the field).
+    /// When empty, a random value is used, which matches the behavior prior to this field existing.
+    pub container_ids: Vec<String>,
+    /// Pool of possible External Data strings for the `|e:` origin detection extension field.
+    ///
+    /// When non-empty, each generated metric randomly selects from this pool (or omits the field).
+    /// Format: `pu-<pod_uid>,cn-<container_name>,it-<bool>`.
+    /// When empty, the field is never emitted.
+    pub external_data: Vec<String>,
+    /// Pool of possible cardinality values for the `|card:` origin detection extension field.
+    ///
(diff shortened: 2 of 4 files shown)


lading_payload/src/common/strings.rs
```diff
@@ -62,6 +62,9 @@ pub(crate) fn choose_or_not_ref<'a, R, T>(mut rng: &mut R, pool: &'a [T]) -> Opt
 where
     R: rand::Rng + ?Sized,
 {
+    if pool.is_empty() {
+        return None;
+    }
```
thieman, line 67: Change to prevent advancing the RNG if the pool is empty. This was failing fingerprint tests because our new external_data and cardinality fields default to empty. This change does not affect any of the other three existing usages of `choose_or_not_ref` because they were never called with empty pools

lading_payload/src/dogstatsd.rs
```diff
@@ -752,6 +784,102 @@ mod test {
...
+            .to_bytes(rng, 65536, &mut bytes)
+            .expect("failed to generate bytes");
+        bytes
+    }
+
+    // With a non-empty pool, the extension field must appear in at least one metric across a
+    // set of differently-seeded runs. Each metric has a 50% chance of including the field, so
+    // the probability of never seeing it across N independent runs is 0.5^N (~10^-30 for N=100).
+    fn assert_field_emitted_at_least_once(config: &Config, field_prefix: &[u8]) {
+        let seen = (0..100).any(|seed| {
```
thieman, line 802: This is technically a flakeable test but if anyone ever sees it fail they should go out and buy a lottery ticket. Can remove it if that's still not okay.

blt: I buy the analysis here, no concerns.

blt (approved):

PR #1873, opened 2026-05-12
Add DogStatsD timestamp extension generation

thieman:
## Summary
- add DogStatsD protocol v1.3 `|T` timestamp generation for count and gauge metrics
- configure timestamps with `timestamp.range` and `timestamp.probability`
- add tests for emission, omission, range bounds, count/gauge-only behavior, and determinism
- update the DogStatsD example and changelog

## Validation
- `ci/fmt` passed
- `cargo check -p lading-payload --tests` passed
- `cargo clippy -p lading-payload --tests -- -D warnings` passed
- `cargo test -p lading-payload dogstatsd --lib` passed
- `ci/validate` attempted, but local full workspace check is blocked by missing system `fuse`/`osxfuse` libraries required by `fuser` (`fuse.pc`/`osxfuse.pc` not found via pkg-config)

### DIFF
--- a/examples/dogstatsd-generation.yaml
+++ b/examples/dogstatsd-generation.yaml
@@ -30,6 +30,21 @@ generator:
             distribution: 0
             set: 0
             histogram: 0
+          container_ids:
+            - "abc123def456"
+            - "def456abc123"
+          external_data:
+            - "pu-abc123,cn-mycontainer,it-false"
+            - "pu-def456,cn-othercontainer,it-true"
+          cardinality:
+            - "low"
+            - "orchestrator"
+          timestamp:
+            range:
+              inclusive:
+                min: 1656581400
+                max: 1656585000
+            probability: 0.25
       bytes_per_second: "80 MiB"
       maximum_prebuild_cache_size_bytes: "1 GiB"
 
--- a/lading_payload/src/dogstatsd.rs
+++ b/lading_payload/src/dogstatsd.rs
@@ -213,6 +213,52 @@ pub struct Config {
     /// When non-empty, each generated metric randomly selects from this pool (or omits the field).
     /// When empty, the field is never emitted.
     pub cardinality: Vec<String>,
+    /// Configuration for the optional `|T` `DogStatsD` protocol v1.3 timestamp field.
+    pub timestamp: Box<TimestampConfig>,
+}
+
+/// Configuration for the optional `|T` `DogStatsD` protocol v1.3 timestamp field.
+#[derive(Debug, Deserialize, serde::Serialize, Clone, Copy, PartialEq)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[serde(deny_unknown_fields, default)]
+pub struct TimestampConfig {
+    /// Range of possible Unix timestamps for the `|T` `DogStatsD` protocol v1.3 field.
+    ///
+    /// The `DogStatsD` protocol only supports this field for count and gauge metrics.
+    pub range: ConfRange<u32>,
+    /// Probability between 0 and 1 that a generated count or gauge metric includes `|T`.
+    pub probability: f32,
+}
+
+impl Default for TimestampConfig {
+    fn default() -> Self {
+        Self {
+            range: ConfRange::Constant(1),
+            probability: 0.0,
+        }
+    }
+}
+
+impl TimestampConfig {
+    fn valid(&self) -> Result<(), String> {
+        let (range_valid, reason) = self.range.valid();
+        if !range_valid {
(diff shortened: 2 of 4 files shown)


examples/dogstatsd-generation.yaml
```diff
@@ -30,6 +30,21 @@ generator:
...
             histogram: 0
+          container_ids:
+            - "abc123def456"
+            - "def456abc123"
+          external_data:
+            - "pu-abc123,cn-mycontainer,it-false"
+            - "pu-def456,cn-othercontainer,it-true"
+          cardinality:
+            - "low"
+            - "orchestrator"
```
thieman, line 41: Added examples for these other extension fields which were implemented in https://github.com/DataDog/lading/pull/1871

lading_payload/src/dogstatsd/metric.rs
```diff
@@ -98,6 +104,22 @@ impl MetricGenerator {
...
+        if self.timestamp_probability == 0.0 {
+            return None;
+        }
+
+        let prob: f32 = OpenClosed01.sample(rng);
+        if prob < self.timestamp_probability {
+            Some(self.timestamp_range.sample(rng))
+        } else {
+            None
+        }
```
goxberry: I suspect you may want to change the distribution you're sampling from `rand::distr::OpenClosed01` to `rand::distr::StandardUniform`.

I would expect `self.timestamp_probability == 1.0f32` to always generate a timestamp. However, it is possible for `rand::distr::OpenClosed01` to sample the value 1.0. In that scenario, `prob < self.timestamp_probability` would evaluate to `false`, so no timestamp would be generated, which would yield counterintuitive behavior.

lading_payload/src/dogstatsd.rs
```diff
@@ -213,6 +213,52 @@ pub struct Config {
     /// When non-empty, each generated metric randomly selects from this pool (or omits the field).
     /// When empty, the field is never emitted.
     pub cardinality: Vec<String>,
+    /// Configuration for the optional `|T` `DogStatsD` protocol v1.3 timestamp field.
+    pub timestamp: Box<TimestampConfig>,
```
goxberry, line 217: Could you help me understand why this field is boxed? `TimestampConfig` isn't that much larger than a `ConfRange<f32>` or a `ConfRange<u32>`, but the existing `Config` struct definition doesn't box fields with those types.

goxberry (approved):
Had some minor comments; other than those, it looks good to me!

lading_payload/src/dogstatsd.rs
thieman: Without the box we hit a Clippy error around the DogStatsD enum variant for Config being too large:

```
   error: large size difference between variants
      --> lading_payload/src/lib.rs:129:1
       |
   129 | /  pub enum Config {
   130 | |      /// Generates Fluent messages
   131 | |      Fluent,
   132 | |      /// Generates syslog5424 messages
   ...   |
   157 | |/     StaticTimestamped {
   158 | ||         /// Defines the file path to read static variant data from. Content is
   159 | ||         /// assumed to be line-oriented.
   160 | ||         static_path: PathBuf,
   ...   ||
   170 | ||         start_line_index: Option<u64>,
   171 | ||     },
       | ||_____- the second-largest variant contains at least 65 bytes
   ...   |
   186 | |      DogStatsD(crate::dogstatsd::Config),
       | |      ----------------------------------- the largest variant contains at least 272 bytes
   ...   |
   195 | |      },
   196 | |  }
       | |__^ the entire enum is at least 272 bytes
       |
       = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
       = note: `-D clippy::large-enum-variant` implied by `-D clippy::all`
       = help: to override `-D clippy::all` add `#[allow(clippy::large_enum_variant)]`
   help: consider boxing the large fields or introducing indirection in some other way to reduce the total size of the enum
       |
   186 -     DogStatsD(crate::dogstatsd::Config),
   186 +     DogStatsD(Box<crate::dogstatsd::Config>),
```

This is more of a bandaid though as it's not consistent throughout the struct as you note. I'll go ahead and merge this as-is since it looks like we don't have any working hours overlap currently, but happy to change this up in a follow-up PR to be consistent if you'd rather.

lading_payload/src/dogstatsd/metric.rs
thieman: Thanks! I think Claude probably got this from these other two usages in the file, it seems like these would suffer from the same problem but wanted to check with you first: https://github.com/DataDog/lading/blob/ac9adbca090b22973f476dbf04289b744e22c6f5/lading_payload/src/dogstatsd/metric.rs#L145-L161

I can fix those in a follow-up if they're bugged.

PR #1875, opened 2026-05-16
fix(dogstatsd): reject tag_length.end() <= MIN_TAG_LENGTH upfront

blt:
## Summary

The dogstatsd `common::tags::Generator` wrapper reserves one byte of the on-wire `tag_length` for the `:` separator before forwarding to the inner `common::tags::Generator`, by building `Inclusive { min: start, max: end - 1 }`. The inner generator requires a non-empty range whose `start() >= MIN_TAG_LENGTH = 3`.

Because only `max` is decremented (not `min`), **any range where `start == end` collapses** — the adjusted range becomes `Inclusive { N, N - 1 }` (`min > max`). That includes `ConfRange::Constant(N)` for any `N`, and `Inclusive { min: N, max: N }`. The inner generator then rejected the range, and the failure bubbled up as `Error::StringGenerate`, which drops the message and points callers at pool generation rather than at the tag-length range that was actually at fault.

This change:

- Generalizes the upfront validation: a `tag_length` is rejected when `end` is not strictly greater than **both** `start` and `MIN_TAG_LENGTH`, via a dedicated `Error::TagLengthRangeTooNarrow` variant that names the offending range. This catches every collapsing case — values at/below the minimum (`Constant(3)`, `Inclusive { 3, 3 }`) **and** constant/single-value ranges above the minimum (`Constant(4)`, `Inclusive { 100, 100 }`).
- Propagates that variant through the public `DogStatsD::new` boundary as `crate::Error::Validation` (message intact), instead of the previous catch-all `StringGenerate`. Other tag-generator failures keep the existing `StringGenerate` behaviour.

Generation semantics are intentionally unchanged (no fingerprint churn): collapsing ranges are *rejected* with a clear error rather than being silently reinterpreted.

Discovered while fuzzing `DogStatsD::new` from an external rig: the symptom was a `StringGenerate` failure at specific seeds, with no indication that the root cause was the configured tag-length range collapsing.

## Test plan

- [x] `cargo test -p lading-payload` — passes, including:
  - wrapper rejection tests: `tag_length_constant_at_min_rejected`, `tag_length_inclusive_min_equals_max_at_min_rejected`, `tag_length_constant_above_min_rejected`, `tag_length_inclusive_single_value_above_min_rejected`;
  - acceptance test for the smallest valid range: `tag_length_at_min_plus_one_accepted`;
  - table-driven public-path test over collapsing configs: `collapsing_tag_length_surfaces_as_validation_error`.
- [x] `cargo clippy -p lading-payload --all-targets` — clean.
- [x] `cargo fmt --check` — clean.
- [x] CHANGELOG entry added under `Unreleased`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   metric, or omits the field entirely if the pool is empty.
 - `dogstatsd` generator now supports DogStatsD protocol v1.3 `|T` timestamps
   for count and gauge metrics via `timestamp.range` and `timestamp.probability`.
+- Fixed: `dogstatsd::common::tags::Generator::new` would silently produce an
+  invalid `Inclusive { min, max: min - 1 }` range when the caller's
+  `tag_length.end()` equalled `MIN_TAG_LENGTH` (e.g. `Constant(3)` or
+  `Inclusive { min: 3, max: 3 }`). The inner generator then rejected the range
+  and the error surfaced as a misleading `Error::StringGenerate` upstream. The
+  wrapper now validates upfront and returns `Error::TagLengthEndTooSmall` with
+  a message identifying the violated constraint.
 
 ## Removed
 - Removed no longer used `smaps.private_hugetlb.by_pathname` procfs observer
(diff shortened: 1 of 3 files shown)


goxberry (approved):
Solid.

PR #1876, opened 2026-05-20
feat(payload): improve BoundedProbability ergonomics for adoption

goxberry:
### What does this PR do?

Prepares `BoundedProbability` (added in #1874) to replace bare `f32`
probability fields throughout `lading_payload` configs. The PR is a
sequence of independently-reviewable commits that add the surface the
adoption stack actually needs, plus a final cleanup commit that drops
the items the adoption PRs turned out not to use:

1. **`impl PartialEq/Eq/PartialOrd/Ord/Hash`** — hand-written, sound under
   `try_new`'s invariants (no NaN, no `-0.0`). `PartialEq`/`Ord` use `f32`
   operators (`==`, `total_cmp`); `Hash` uses the `u32` bit pattern because
   `f32: !Hash`. The `Eq`/`Hash` contract holds because numerically equal
   valid values share a bit pattern.
2. **Type aliases** — rename the generic struct to
   `BoundedProbability<MIN_AS_BITS>` and add `Probability` (`[0.0, 1.0]`)
   and `AtLeastOneTenth` (`[0.1, 1.0]`) so callers don't have to spell
   `f32::to_bits(...)` at each use site.
3. **`impl arbitrary::Arbitrary`** (feature-gated) — needed so structs
   containing a `Probability` field can keep deriving `Arbitrary` once
   wired in. A derived impl would emit NaN / infinity / out-of-range
   values; this one samples a `u32` in `[MIN_AS_BITS, to_bits(+1.0)]` and
   relies on the documented `f32` <-> `u32` monotonicity to guarantee
   validity.
4. **`sample_bernoulli` helper** — replaces the
   `rng.random_bool(p.get() as f64)` pattern that every adopting call site
   would otherwise repeat. The `f32` -> `f64` conversion is exact on
   `[0.0, 1.0]`, so the success probability is preserved bit-for-bit.
5. **`AtLeastOneHundredth` alias** — `BoundedProbability<{ f32::to_bits(0.01) }>`
   for `unique_tag_ratio`'s existing 0.01 floor, which must admit
   in-the-wild values below 0.1. Also drops the `unique_tag_ratio`
   reference from the `AtLeastOneTenth` rustdoc.
6. **Cleanup** — drop the items the adoption stack (#1877–#1881) does
   not exercise: `impl Display` (originally from #1874), `impl Eq`,
   `impl Ord`, `impl PartialOrd`, `impl Hash`, `sample_bernoulli`, and
   the `AtLeastOneTenth` alias. Surviving surface: `PartialEq`
   (transitively required by `dogstatsd::Config`/`ValueConf`/`TimestampConfig`
   deriving `PartialEq`), `Arbitrary` (transitively required by fuzz
   targets that derive `Arbitrary` on `Config`), `TryFrom<f32>` /
   `From<BoundedProbability<_>> for f32` (required by
   `serde(into = "f32", try_from = "f32")`), `try_new`, `get`, and the
   `Probability` and `AtLeastOneHundredth` aliases. The dead tests are
   pruned alongside.

Each non-cleanup commit ships with unit and property tests; the cleanup
commit removes the now-dead tests covering removed items.

### Motivation

`BoundedProbability` exists but isn't yet usable in `Config` structs:
adopters need `PartialEq` to keep deriving that trait, `Arbitrary` for
fuzz harnesses, and named aliases instead of bit-pattern incantations.
Commits 1–5 build out a wider surface area than needed; commit 6 trims
back to what the rest of the stack actually calls. The history is
preserved so the rationale for each piece is reviewable, with the
cleanup commit acting as a journaled "what stuck" record.

### Related issues

Follow-up to #1874. Foundation for the adoption stack:
- #1877 — `TimestampConfig::probability`
- #1878 — `ValueConf::float_probability`
- #1879 — `Config::sampling_probability`
- #1880 — `Config::multivalue_pack_probability`
- #1881 — `Config::unique_tag_ratio` (`AtLeastOneHundredth`)

### Additional Notes

No wire-format change — `BoundedProbability` still serializes as a bare
`f32` via `serde(into = "f32", try_from = "f32")`. Phase 1 PRs in the
stack switch one `Config` field per PR with no behavioral change at
runtime (`.get()` inlines to the same machine code as direct `f32`
access, and no sampler is swapped).

### DIFF
--- a/lading_payload/src/common/config.rs
+++ b/lading_payload/src/common/config.rs
@@ -293,6 +293,26 @@ impl<const MIN_AS_BITS: u32> BoundedProbability<MIN_AS_BITS> {
     }
 }
 
+/// Generate a uniformly-distributed-over-bit-patterns value in `[MIN, +1.0]`
+/// by sampling a `u32` in `[MIN_AS_BITS, f32::to_bits(+1.0)]` and decoding it.
+///
+/// This works because the f32 ↔ u32 ordering (documented on the type) is
+/// monotonic for non-negative finite values, so every bit pattern in that
+/// range decodes to a valid stored value. `-0.0`'s bit pattern is
+/// `0x8000_0000`, far above `f32::to_bits(+1.0) = 0x3f80_0000`, so it can
+/// never be generated.
+#[cfg(feature = "arbitrary")]
+impl<'a, const MIN_AS_BITS: u32> arbitrary::Arbitrary<'a> for BoundedProbability<MIN_AS_BITS> {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        let bits = u.int_in_range(MIN_AS_BITS..=f32::to_bits(Self::MAX))?;
+        let value = f32::from_bits(bits);
+        // Routing through `try_new` fires the per-monomorphization const-eval
+        // bound check on `Self::MIN` and forwards any future invariant added
+        // to the constructor. The `expect` is safe by the argument above.
+        Ok(Self::try_new(value).expect("bits in [MIN_AS_BITS, MAX_AS_BITS] always valid"))
+    }
+}
+
 #[cfg(test)]
 mod probability_tests {
     use super::{BoundedProbability, NEG_ZERO_AS_BITS, ProbabilityError};
@@ -737,6 +757,47 @@ mod probability_tests {
                 err.to_string().contains("exceeds 1.0"),
                 "unexpected error: {}", err
             );
+        }
+    }
+
+    // ===== Property tests: Arbitrary impl (feature = "arbitrary") =====
+
+    #[cfg(feature = "arbitrary")]
+    fn check_arbitrary_produces_valid<const MIN_AS_BITS: u32>(bytes: &[u8]) {
+        use arbitrary::{Arbitrary, Unstructured};
+        let mut u = Unstructured::new(bytes);
+        // `int_in_range` can fail with `NotEnoughData` on short inputs; that's
+        // fine — we only need to check that any `Ok` value is valid.
+        if let Ok(p) = BoundedProbability::<MIN_AS_BITS>::arbitrary(&mut u) {
+            let v = p.get();
+            assert!(v.is_finite());
+            assert_ne!(v.to_bits(), NEG_ZERO_AS_BITS);
+            assert!(v >= BoundedProbability::<MIN_AS_BITS>::MIN);
+            assert!(v <= BoundedProbability::<MIN_AS_BITS>::MAX);
+        }
+    }
+
+    #[cfg(feature = "arbitrary")]
+    proptest! {
+        #[test]
+        fn arbitrary_produces_valid_zero_or_more(
+            bytes in prop::collection::vec(any::<u8>(), 4..32),
+        ) {
+            check_arbitrary_produces_valid::<{ f32::to_bits(0.0) }>(&bytes);
+        }
+
+        #[test]
+        fn arbitrary_produces_valid_at_least_half(
+            bytes in prop::collection::vec(any::<u8>(), 4..32),
+        ) {
+            check_arbitrary_produces_valid::<{ f32::to_bits(0.5) }>(&bytes);
+        }
+
+        #[test]
+        fn arbitrary_produces_valid_at_least_one(
+            bytes in prop::collection::vec(any::<u8>(), 4..32),
+        ) {
+            check_arbitrary_produces_valid::<{ f32::to_bits(1.0) }>(&bytes);
         }
     }
 }
--- a/lading_payload/src/common/config.rs
+++ b/lading_payload/src/common/config.rs
@@ -2,7 +2,7 @@
 
 use rand::{RngExt, distr::uniform::SampleUniform};
 use serde::Deserialize;
-use std::{cmp, fmt, hash};
+use std::{cmp, fmt};
 
 /// Range expression for configuration
 #[derive(Debug, Deserialize, serde::Serialize, Clone, PartialEq, Copy)]
@@ -174,10 +174,6 @@ pub struct BoundedProbability<const MIN_AS_BITS: u32> {
 /// A probability in the closed unit interval `[0.0, 1.0]`. The most common bound.
 pub type Probability = BoundedProbability<{ f32::to_bits(0.0) }>;
 
-/// A probability or ratio in `[0.1, 1.0]`. Use for fields that must avoid
-/// extreme low values.
-pub type AtLeastOneTenth = BoundedProbability<{ f32::to_bits(0.1) }>;
-
 /// A probability or ratio in `[0.01, 1.0]`. Use for fields such as
 /// `unique_tag_ratio` that must avoid extreme low values but admit
 /// in-the-wild values below `0.1`.
@@ -197,41 +193,9 @@ impl<const MIN_AS_BITS: u32> From<BoundedProbability<MIN_AS_BITS>> for f32 {
     }
 }
 
-impl<const MIN_AS_BITS: u32> fmt::Display for BoundedProbability<MIN_AS_BITS> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        fmt::Display::fmt(&self.value, f)
-    }
-}
-
-// `Eq`, `Ord`, and `Hash` are sound because [`Self::try_new`] rejects NaN and
-// normalizes `-0.0` to `+0.0`, so every value in the valid range has a unique
-// bit pattern and an unambiguous numeric ordering. `Hash` works on the `u32`
-// bit pattern because `f32: !Hash`; the `Eq`/`Hash` contract is preserved
-// because numerically equal valid values share a bit pattern.
-
 impl<const MIN_AS_BITS: u32> PartialEq for BoundedProbability<MIN_AS_BITS> {
     fn eq(&self, other: &Self) -> bool {
         self.value == other.value
-    }
-}
-
-impl<const MIN_AS_BITS: u32> Eq for BoundedProbability<MIN_AS_BITS> {}
-
-impl<const MIN_AS_BITS: u32> Ord for BoundedProbability<MIN_AS_BITS> {
-    fn cmp(&self, other: &Self) -> cmp::Ordering {
-        self.value.total_cmp(&other.value)
-    }
-}
-
-impl<const MIN_AS_BITS: u32> PartialOrd for BoundedProbability<MIN_AS_BITS> {
-    fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
-        Some(self.cmp(other))
-    }
-}
-
-impl<const MIN_AS_BITS: u32> hash::Hash for BoundedProbability<MIN_AS_BITS> {
-    fn hash<H: hash::Hasher>(&self, state: &mut H) {
-        self.value.to_bits().hash(state);
     }
 }
 
@@ -295,19 +259,6 @@ impl<const MIN_AS_BITS: u32> BoundedProbability<MIN_AS_BITS> {
     #[must_use]
     pub const fn get(&self) -> f32 {
         self.value
-    }
-
-    /// Sample a Bernoulli trial with success probability `self.get()`.
-    ///
-    /// Returns `true` with probability `self.get()` and `false` otherwise.
-    /// The f32 <-> f64 conversion is exact for every f32 in `[+0.0, +1.0]`, so
-    /// the success probability is preserved bit-for-bit.
-    #[must_use]
-    pub fn sample_bernoulli<R>(&self, rng: &mut R) -> bool
-    where
-        R: rand::Rng + ?Sized,
-    {
-        rng.random_bool(f64::from(self.value))
     }
 }
 
@@ -394,7 +345,7 @@ mod probability_tests {
         }
     }
 
-    // ===== Unit tests: ordering / equality / hashing =====
+    // ===== Unit tests: equality =====
 
     #[test]
     fn equality_holds_for_same_bit_pattern() {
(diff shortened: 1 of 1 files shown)


goxberry:
* **#1881** <a href="https://app.graphite.com/github/pr/DataDog/lading/1881?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1880** <a href="https://app.graphite.com/github/pr/DataDog/lading/1880?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1879** <a href="https://app.graphite.com/github/pr/DataDog/lading/1879?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1878** <a href="https://app.graphite.com/github/pr/DataDog/lading/1878?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1877** <a href="https://app.graphite.com/github/pr/DataDog/lading/1877?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1876** <a href="https://app.graphite.com/github/pr/DataDog/lading/1876?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/lading/1876?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

goxberry:
# BoundedProbability rollout plan

Wire the `BoundedProbability` aliases (`Probability`, `AtLeastOneTenth`, and a
new `AtLeastOneHundredth`) into the `lading_payload` config and generator
types. Keep `Generator::generate` paths bit-exact and zero-overhead.

## Decisions locked in
- `unique_tag_ratio`: add `AtLeastOneHundredth = BoundedProbability<{ f32::to_bits(0.01) }>`
  to preserve the existing 0.01 floor. Do NOT use `AtLeastOneTenth` for this
  field (it would tighten the bound and reject in-the-wild configs in
  `[0.01, 0.10)`).
- PR structure: PR the existing branch first; then one stacked PR per field.
- Stack tool: Graphite (`gt`). Repo initialized with `main` as trunk;
  `goxberry/improve-probability-type-ergonomics` tracked as child of `main`.

## Phase 0 — Land the existing branch
On `goxberry/improve-probability-type-ergonomics`, add:
1. `AtLeastOneHundredth = BoundedProbability<{ f32::to_bits(0.01) }>` alias
   in `lading_payload/src/common/config.rs`.
2. Update the rustdoc on `AtLeastOneTenth` so it does NOT claim
   `unique_tag_ratio` uses it (since we're keeping `unique_tag_ratio` on
   the 0.01 floor).

Then `gt submit` to push and open the PR.

## Phase 1 — Stacked PRs (one per field)

### Per-field recipe
- Change the public `Config` / `TimestampConfig` / `ValueConf` field type from
  `f32` to the matching alias (`Probability` or `AtLeastOneHundredth`).
- Remove the now-redundant `is_finite() && (0.0..=1.0).contains(&x)` checks
  from `valid()` — `try_from` enforces them at deserialize time.
- Thread the alias type down through the private constructors
  (`MemberGenerator::new`, `MetricGenerator::new`, `NumValueGenerator::new`,
  `common::tags::Generator::new`, `dogstatsd::common::tags::Generator::new`,
  `opentelemetry::common::TagGenerator::new`).
- At the comparison site in the generate hot path, call `.get()` to extract
  the `f32`. `.get()` is `const fn` returning a copied scalar — inlines to
  the same machine code, so RNG sequences and bit-exact behavior are
  preserved.
- Update test fixtures and `Default` impls that construct these structs
  literally.

### Hot-path call sites that stay `f32` at comparison time
- `lading_payload/src/dogstatsd/metric.rs:112` `self.timestamp_probability == 0.0`
- `lading_payload/src/dogstatsd/metric.rs:117` `prob < self.timestamp_probability`
- `lading_payload/src/dogstatsd/metric.rs:147` `prob < self.sampling_probability`
- `lading_payload/src/dogstatsd/metric.rs:161` `prob < self.multivalue_pack_probability`
- `lading_payload/src/dogstatsd/common.rs:70` and `:81` `prob < *float_probability`
- `lading_payload/src/common/tags.rs:277` `choose_existing_prob > self.unique_tag_probability`

All become `.get()` calls. Do NOT substitute `sample_bernoulli` —
`OpenClosed01 < p` and `random_bool(p)` are not bit-identical samplers, and
swapping would shift the user-visible RNG sequence.

### Stack order (smallest blast radius first)

| PR | Field | Alias | Notes |
|---|---|---|---|
| 1 | `TimestampConfig::probability` | `Probability` | Self-contained struct with its own `valid()`. Many test fixtures in `dogstatsd.rs` construct `TimestampConfig` literally. |
| 2 | `ValueConf::float_probability` | `Probability` | Threads into `NumValueGenerator`. |
| 3 | `Config::sampling_probability` | `Probability` | Threads through `MemberGenerator::new` -> `MetricGenerator::new`. |
| 4 | `Config::multivalue_pack_probability` | `Probability` | Same path as PR 3. |
| 5 | `Config::unique_tag_ratio` | `AtLeastOneHundredth` | Threads through `MemberGenerator::new` -> `common::tags::Generator::new` (also `dogstatsd::common::tags::Generator::new` wrapper and `opentelemetry::common::TagGenerator::new`). The `opentelemetry::common::UNIQUE_TAG_RATIO: f32 = 0.75` const becomes a `const AtLeastOneHundredth` via `match try_new(0.75)`. Highest blast radius — tag tests reference the existing const bounds. |

### Efficiency verification
After each Phase 1 PR, run the relevant criterion benches and compare against
the prior commit:
- PRs 1–4: `cargo bench -p lading_payload --bench dogstatsd` (and `block`).
- PR 5: also OTel benches (`opentelemetry_log`, `opentelemetry_metric`,
  `opentelemetry_traces`).

Expected delta is noise — no allocation change, no new branch, no extra
function call (the `.get()` inlines away).

## Out of scope
- No change to YAML/JSON wire format. `BoundedProbability` continues to
  serialize as a bare `f32` via `serde(into = "f32", try_from = "f32")`.
- No swap of `OpenClosed01 < p` for `sample_bernoulli` anywhere.
- No change to the existing `common::tags::MIN_UNIQUE_TAG_RATIO`,
  `WARN_UNIQUE_TAG_RATIO`, or `MAX_UNIQUE_TAG_RATIO` constants — kept for
  backward compatibility and reuse in tests.

lading_payload/src/common/config.rs
```diff
@@ -147,42 +147,95 @@ pub enum ProbabilityError {
...
+    fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl<const MIN_AS_BITS: u32> hash::Hash for BoundedProbability<MIN_AS_BITS> {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
+        self.value.to_bits().hash(state);
+    }
+}
```
goxberry: I suspect this code is unnecessary and should be deleted.

lading_payload/src/common/config.rs
```diff
@@ -243,16 +296,49 @@ impl<const MIN_AS_BITS: u32> Probability<MIN_AS_BITS> {
...
+    /// The f32 <-> f64 conversion is exact for every f32 in `[+0.0, +1.0]`, so
+    /// the success probability is preserved bit-for-bit.
+    #[must_use]
+    pub fn sample_bernoulli<R>(&self, rng: &mut R) -> bool
+    where
+        R: rand::Rng + ?Sized,
+    {
+        rng.random_bool(f64::from(self.value))
+    }
+}
```
goxberry: I'm not sure this code is worth keeping either. It isn't used elsewhere in the stack. `random_bool` creates a `rand_distr::Bernoulli(self.value)` instance every time it's called, which seems like a drawback in scenarios where we'd repeatedly sample from the same distribution.

lading_payload/src/common/config.rs
```diff
@@ -243,16 +296,49 @@ impl<const MIN_AS_BITS: u32> Probability<MIN_AS_BITS> {
...
+#[cfg(feature = "arbitrary")]
+impl<'a, const MIN_AS_BITS: u32> arbitrary::Arbitrary<'a> for BoundedProbability<MIN_AS_BITS> {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        let bits = u.int_in_range(MIN_AS_BITS..=f32::to_bits(Self::MAX))?;
+        let value = f32::from_bits(bits);
+        // Routing through `try_new` fires the per-monomorphization const-eval
+        // bound check on `Self::MIN` and forwards any future invariant added
+        // to the constructor. The `expect` is safe by the argument above.
+        Ok(Self::try_new(value).expect("bits in [MIN_AS_BITS, MAX_AS_BITS] always valid"))
+    }
```
goxberry, line 282: I don't think this trait implementation would actually get used unless we were fuzzing this type for some reason. I can't think of a scenario where we would actually do that. I think this block of code could be deleted without any real consequence.

goxberry:
# Remove unused `BoundedProbability` trait impls, methods, and aliases

## Context

PR 1876 (`goxberry/improve-probability-type-ergonomics`, current branch) is the
foundation of a five-PR stack that extends `BoundedProbability` and then adopts
it across `lading_payload`. After the adoption PRs (1877–1881) land at the
stack tip (`goxberry/probability-unique-tag-ratio`), several items added in
PR 1876 turn out to have no production caller. Carrying them forward enlarges
the API surface, the test surface, and the supported invariants without any
caller benefiting.

The user wants PR 1876 trimmed so the type only exposes what the rest of the
stack actually uses. Audit performed against the tip of the stack
(`goxberry/probability-unique-tag-ratio`) — anything not referenced outside
`lading_payload/src/common/config.rs` is removable.

## Audit summary (against tip of PR 1881)

Production / transitive callers verified via `git grep` against
`goxberry/probability-unique-tag-ratio`:

**Keep (callers exist):**
- `BoundedProbability` type, `Probability` alias, `AtLeastOneHundredth` alias
- `try_new`, `get` methods
- `TryFrom<f32>`, `From<BoundedProbability<_>> for f32` — required by
  `#[serde(into = "f32", try_from = "f32")]`
- `impl PartialEq` — transitively required because `dogstatsd::Config`,
  `dogstatsd::ValueConf`, `dogstatsd::TimestampConfig` derive `PartialEq` and
  hold `Probability` / `AtLeastOneHundredth` fields
- `impl Arbitrary` (feature = "arbitrary") — transitively required by fuzz
  targets in `lading_payload/fuzz/fuzz_targets/dogstatsd_*.rs` that derive
  `arbitrary::Arbitrary` on `Input { config: lading_payload::dogstatsd::Config, .. }`
- `ProbabilityError`, `NEG_ZERO_AS_BITS`, `MIN`, `MAX` — internal callers in
  `try_new` and the `Arbitrary` impl

**Remove (no production caller):**
- `impl fmt::Display for BoundedProbability` — only tests format the type
- `impl Eq` — never required by a derive in production
- `impl Ord` — never required by a derive in production
- `impl PartialOrd` — never required by a derive in production
- `impl hash::Hash` — never required by a derive in production
- `BoundedProbability::sample_bernoulli` — no caller; only tested in-place
- `AtLeastOneTenth` alias — no caller

## Critical file

`lading_payload/src/common/config.rs` is the only file touched.

## Edits in `lading_payload/src/common/config.rs`

**Module-level imports (line 5):**
- Change `use std::{cmp, fmt, hash};` to `use std::{cmp, fmt};`. `cmp` and
  `fmt` are still used by `ConfRange` (`PartialOrd` bounds, `Display` impl).
  `hash` is no longer needed.

**Type aliases (around lines 177–179):**
- Delete the `AtLeastOneTenth` alias and its doc comment.

**Trait impls on `BoundedProbability` (lines 200–236):**
- Delete `impl<const MIN_AS_BITS: u32> fmt::Display for BoundedProbability<MIN_AS_BITS>` block.
- Delete the explanatory comment block on lines 206–210 about Eq/Ord/Hash soundness.
- Delete `impl<const MIN_AS_BITS: u32> Eq for BoundedProbability<MIN_AS_BITS> {}`.
- Delete `impl<const MIN_AS_BITS: u32> Ord for BoundedProbability<MIN_AS_BITS>` block.
- Delete `impl<const MIN_AS_BITS: u32> PartialOrd for BoundedProbability<MIN_AS_BITS>` block.
- Delete `impl<const MIN_AS_BITS: u32> hash::Hash for BoundedProbability<MIN_AS_BITS>` block.
- Keep the `PartialEq` impl above the comment block.

**Inherent methods on `BoundedProbability` (lines 300–311):**
- Delete the `sample_bernoulli` method and its doc comment.

**Tests in `mod probability_tests` (lines 334–850):**
- Remove `ordering_matches_numeric_ordering` (lines 408–416) — uses `<`.
- Remove `hash_agrees_with_eq` (lines 418–427) — uses `HashSet`/`Hash`.
- Remove `check_display_matches`, `check_display_precision` helpers (lines 499–507).
- Remove the three `display_matches_inner_f32_for_valid_values_*` proptests.
- Remove the three `display_propagates_precision_*` proptests.
- Remove `ord_matches_f32_partial_cmp` proptest (lines 605–618) — uses `cmp`.
- Remove `bernoulli_at_zero_never_succeeds` and `bernoulli_at_one_always_succeeds`
  proptests (lines 829–848) and the `// ===== Property tests: sample_bernoulli =====`
  comment block above them.
- Keep `equality_holds_for_same_bit_pattern` — relies only on `PartialEq`.

No other files need to change. The adoption PRs do not reference any of the
removed items, so the stack should rebase through cleanly.

## Verification

Run from the repo root, on the current branch
(`goxberry/improve-probability-type-ergonomics`):

1. `cargo test -p lading-payload` — full test suite passes (previously 250
   tests on PR 1881; expect a lower count after removing tests, all green).
2. `cargo clippy -p lading-payload --all-targets -- -D warnings` — clean
   (catches any orphan `use` or dead-code warnings from removed helpers).
3. `cargo build -p lading-payload --features arbitrary` — confirms the
   `Arbitrary` impl path still compiles after the test changes.
4. Optional smoke: rebase a descendant branch (e.g. `goxberry/probability-timestamp-config`)
   onto this commit locally and run `cargo check -p lading-payload`.
   No expected fallout — none of the adoption PRs use the removed items.

lading_payload/src/common/config.rs
goxberry: Deleted in https://github.com/DataDog/lading/pull/1876/commits/bce8ac61e83184c56775f64a191cccff92a3bbeb.

lading_payload/src/common/config.rs
goxberry: Deleted in https://github.com/DataDog/lading/pull/1876/commits/bce8ac61e83184c56775f64a191cccff92a3bbeb.

blt (approved):

PR #1895, opened 2026-05-23
Add generated OpenMetrics scrape bodies to HTTP blackhole

scottopell:
## Summary
Adds a generated OpenMetrics scrape body for the HTTP blackhole, so scrape-based tests can serve realistic metrics without embedding static response text.

## Changed
- Added a configurable OpenMetrics generator for counters, gauges, histograms, summaries, target metadata, and build metadata.
- OpenMetrics bodies are precomputed at startup and served as static bytes on the HTTP response path.
- OpenMetrics response headers now derive from the body variant, so selecting the body automatically serves the matching OpenMetrics content type.
- Added validation for metric-name prefixes, bucket and quantile bounds, duplicate quantiles, enabled TYPE metadata, and disabled-family edge cases.
- Ensured generated series remain unique when configured counts exceed label-pool cardinality by adding deterministic per-series labels.
- Preserved existing tagged static-body YAML while also accepting singleton-map body variants for generated OpenMetrics config.

## Why
This gives scrape tests a deterministic, spec-shaped OpenMetrics target while keeping HTTP blackhole configuration concise and backward-compatible.


blt (commented):
This is looking good. @scottopell give me a ping when it's open for review and I'll have a look.

scottopell:
@blt the bots are happy, ready for a human stamp

blt (approved):

PR #1905, opened 2026-06-01
Split TCP RR infrastructure from the workload

usamasaqib:
### What does this PR do?

This PR refactors the tcp_rr workload to split the infrastructure elements into a separate file. This will allow subsequent workloads to reuse the same thread and flow management code.

### Additional Notes
PR stack
[1] https://github.com/DataDog/lading/pull/1905 <-- This
[2] https://github.com/DataDog/lading/pull/1906


lading/src/neper/flow.rs
```diff
@@ -73,12 +73,16 @@ pub(crate) fn apply_action<S>(
         Action::Continue => {}
         Action::Reregister(interest) => {
             if let Some(flow) = flows.get_mut(token) {
-                let _ = registry.reregister(&mut flow.stream, flow.token, interest);
+                registry
+                    .reregister(&mut flow.stream, flow.token, interest)
+                    .expect("reregister of a live, owned flow must succeed");
```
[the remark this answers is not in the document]
usamasaqib: We do not consider this a recoverable error. A live flow must always allow reregistration.

lading/src/generator/tcp_rr.rs
```diff
@@ -54,6 +48,10 @@ fn default_data_port() -> u16 {
 #[derive(Debug, Deserialize, Serialize, PartialEq, Clone)]
 #[serde(deny_unknown_fields)]
 /// Configuration for the `tcp_rr` generator.
+///
+/// Flow count is *not* configured here — it is owned by the
+/// `tcp_rr` blackhole and communicated to the generator over the control port
+/// during startup.
```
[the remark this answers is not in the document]
usamasaqib: It is fine. We do not have many users for this.

lading/src/neper/flow.rs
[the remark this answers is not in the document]
preinlein: Could we bubble up the error instead? I see that this function is called from the tcp_rr (blackhole+generator), we could make `aply_action` return a result and if it's an error, we log and exit lading gracefully.

lading/src/neper/rr.rs
```diff
@@ -0,0 +1,735 @@
...
+                format!(
+                    "shutdown before blackhole control port {} became reachable",
+                    params.control_addr
+                ),
+            )));
+        }
+        match net::TcpStream::connect(params.control_addr) {
+            Ok(mut conn) => {
+                conn.set_read_timeout(Some(HANDSHAKE_TIMEOUT))
+                    .expect("set_read_timeout on connected TcpStream must succeed");
```
preinlein: rather than expect here, could we bubble up the error?

lading/src/neper/rr.rs
```diff
@@ -0,0 +1,735 @@
...
+    info!("shutdown signal received");
+    shutdown_flag.store(true, Relaxed);
+
+    worker_handles.push(metrics_handle);
+    thread::join_all(worker_handles).map_err(|()| Error::ThreadPanicked)?;
+
+    Ok(())
+}
+
+fn client_thread_main(
```
preinlein, line 367: can we make this function bubble up errors as well

lading/src/neper/rr.rs
```diff
@@ -0,0 +1,735 @@
...
+                        metrics.requests_sent.add(1);
+                        metrics.bytes_written.add(request_buf.len() as u64);
+                        Action::Reregister(Interest::READABLE)
+                    } else {
+                        Action::Continue
+                    }
+                }
+                Err(e) if e.kind() == ErrorKind::WouldBlock => Action::Continue,
+                Err(e) => {
+                    trace!("write error: {e}");
```
preinlein, line 452: can we make this an error log

lading/src/neper/rr.rs
```diff
@@ -0,0 +1,735 @@
...
+                        metrics.responses_received.add(1);
+                        metrics.bytes_read.add(response_buf.len() as u64);
+                        Action::Reregister(Interest::WRITABLE)
+                    } else {
+                        Action::Continue
+                    }
+                }
+                Err(e) if e.kind() == ErrorKind::WouldBlock => Action::Continue,
+                Err(e) => {
+                    trace!("read error: {e}");
```
preinlein, line 475: same thing here

lading/src/neper/rr.rs
```diff
@@ -0,0 +1,735 @@
...
+    // signaling (thread panicked), recv() eventually returns None.
+    for _ in 0..num_threads {
+        if ready_rx.recv().await.is_none() {
+            shutdown_flag.store(true, Relaxed);
+            thread::join_all(handles).map_err(|()| Error::ThreadPanicked)?;
+            return Err(Error::ThreadPanicked);
+        }
+    }
+
+    // All data listeners are up. Open control port so the generator can
```
preinlein: I wonder if we can refactor this method so that we don't need the clippy/lint exception for "method too long" here. Can we refactor up to this point into a function called `prepare_data_listeners` or something akin to that?

lading/src/neper/rr.rs
```diff
@@ -0,0 +1,735 @@
...
+        addr: control_addr,
+        source: Box::new(source),
+    });
+    if ctrl_res.is_err() {
+        shutdown_flag.store(true, Relaxed);
+    }
+    let control_listener = ctrl_res?;
+    control_listener
+        .set_nonblocking(true)
+        .expect("failed to set control listener nonblocking");
```
preinlein: Let's do a pass thru the PR and remove all these expects and instead change it to bubble up errors.

lading/src/neper/rr.rs
```diff
@@ -0,0 +1,735 @@
...
+                return Err(Error::Bind {
+                    addr: control_addr,
+                    source: Box::new(e),
+                });
+            }
+        }
+    }
+    drop(control_listener);
+
+    if generator_connected {
```
preinlein: It feels like everything above this could be another function called `wait_for_generator` or something along those lines.

lading/src/neper/rr.rs
```diff
@@ -0,0 +1,735 @@
...
+            }
+        }
+    }
+}
+
+/// Set `TCP_NODELAY` on a mio [`TcpStream`] via a borrowed `socket2::SockRef`.
+fn set_nodelay_mio(stream: &TcpStream, no_delay: bool) {
+    let sock = socket2::SockRef::from(stream);
+    if let Err(e) = sock.set_tcp_nodelay(no_delay) {
+        trace!("failed to set TCP_NODELAY: {e}");
```
preinlein, line 923: is this problematic? Are we ok with a silent error here?

lading/src/neper/rr.rs
```diff
@@ -0,0 +1,735 @@
...
+                        metrics.requests_received.add(1);
+                        metrics.bytes_received.add(request_buf.len() as u64);
+                        Action::Reregister(Interest::WRITABLE)
+                    } else {
+                        Action::Continue
+                    }
+                }
+                Err(e) if e.kind() == ErrorKind::WouldBlock => Action::Continue,
+                Err(e) => {
+                    trace!("read error: {e}");
```
preinlein, line 952: Let's change all these logs in error handlers to either warn/error

preinlein (commented):
Is there any way to include some tests for things in here? 

It's hard to review with confidence given that I'm not familiar enough with the inner workings of neper/etc.

lading/src/neper/rr.rs
usamasaqib: All of the `trace!` are changed to different log levels in the next PR in this stack

lading/src/neper/rr.rs
usamasaqib: Its been changed to a warn!. It is okay for it to be just a log for now since this feature is not available on all kernel versions, and I am not sure how to best handle it yet. So just logging on continuing seems fine for now.

lading/src/neper/rr.rs
usamasaqib: The expects are there because they are meant to be invariants. If these operations do not succeed then the workload will not behave as designed. So this operation must always suceed.

lading/src/neper/rr.rs
usamasaqib: resolved in this next PR

lading/src/neper/rr.rs
usamasaqib: resolved in the next PR

lading/src/neper/rr.rs
preinlein: And why do we need a panic / how does graceful error handling not work here?

We can just have lading exit gracefully with an error, no?

usamasaqib: We can. I'll make the changes.

lading/src/neper/rr.rs
preinlein: It can totally just be a log and not an error that gets bubbled up. If we do that, let's just add a comment saying why it's ok to have a silent error and if there any consequences for this not working as expected (or being available like you mentioned)

lading/src/neper/rr.rs
preinlein: sounds good, thanks!

lading/src/neper/rr.rs
usamasaqib: Done in 64a17d8f8653296bd0fc7cc6f5ccadc5c2df1869

lading/src/neper/rr.rs
usamasaqib: Done in 64a17d8f8653296bd0fc7cc6f5ccadc5c2df1869

usamasaqib:
All the comments about bubbling up the errors are addressed in 
https://github.com/DataDog/lading/pull/1928

lading/src/neper/rr.rs
```diff
@@ -0,0 +1,798 @@
...
+///
+/// # Errors
+///
+/// Returns the underlying `io::Error` from `connect`, `set_read_timeout` or
+/// `read_exact`.
+fn try_control_handshake(control_addr: SocketAddr) -> io::Result<u16> {
+    let mut conn = net::TcpStream::connect(control_addr)?;
+    conn.set_read_timeout(Some(HANDSHAKE_TIMEOUT))?;
+    let mut buf = [0u8; HANDSHAKE_LEN];
+    conn.read_exact(&mut buf)?;
```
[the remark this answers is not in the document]
usamasaqib: We do not need to worry about this case since both the generator and blackhole are written to be well behaved.

preinlein (approved):
LGTM. The expects are refactored/fixed in https://github.com/DataDog/lading/pull/1928.

Since it seems that `lading`'s main has an outstanding bug for graceful exiting on blackhole errors, we'll approve this one so that we can move forward. `expect` is a hard signal but it bypasses that bug. 

Once we address `lading`'s main bug, we should be able to approve the other PR (1928) and we'll have clearer handling. I do have a feeling we're going to need to do a proper pass to remove `expect`s from the code but that's a bigger change and outside the scope of this PR.

PR #1882, opened 2026-05-22
chore(clippy): forbid .expect() in production code

goxberry:
### What does this PR do?

Lands the *mechanism* that deterministically enforces no `.expect()` calls in production code, mirroring the existing `clippy::unwrap_used` policy.

Concretely:

- `Cargo.toml`: adds `expect_used = "deny"` to `[workspace.lints.clippy]`, next to the existing `unwrap_used = "deny"`.
- `clippy.toml`: adds `allow-expect-in-tests = true`, alongside the existing `allow-unwrap-in-tests`. `.expect()` inside `#[cfg(test)]` modules and `#[test]` functions remains permitted.
- Crate-root quarantines: a transitional `#![allow(clippy::expect_used)]` is added at the crate root of each of the four library crates that already contain production `.expect()` sites:
  - `lading/src/lib.rs`, `lading/src/bin/lading.rs`, `lading/src/bin/captool/main.rs`, `lading/src/bin/payloadtool.rs`
  - `lading_capture/src/lib.rs`, `lading_capture/src/bin/fuzz_capture_harness.rs`
  - `lading_payload/src/lib.rs`
  - `lading_throttle/src/lib.rs`
- Permanent bench allows: `#![allow(clippy::expect_used)]` is added to each of the 14 files in `lading_payload/benches/`. Each bench is a separate crate root, and clippy has no `allow-expect-in-benches` equivalent of `allow-expect-in-tests`. The workspace's `warnings = "deny"` (under `[workspace.lints.rust]`) promotes any clippy warning to a hard error, so a per-bench allow is the minimal targeted fix that keeps full clippy coverage on the rest of the bench code.
- `CHANGELOG.md`: entry added under `## Unreleased`.

The gate fires immediately on any new `.expect()` outside `#[cfg(test)]` in any non-quarantined location. New `.expect()` calls added inside the quarantined crates will not fire — that visibility returns crate-by-crate as cleanup PRs land.

### Motivation

`.unwrap()` was already gated at `deny` workspace-wide, but `.expect()` was not — leaving an easy way to introduce production panics. A targeted survey of the workspace turned up **602** production `.expect()` sites across four library crates:

| Crate | Production `.expect()` sites |
| --- | --- |
| `lading` | 243 |
| `lading_payload` | 164 |
| `lading_capture` | 149 |
| `lading_throttle` | 28 |
| `lading_capture_schema` / `lading_signal` / `lading_fuzz` | 0 |

[...]

PR #1906, opened 2026-06-02
TCP CRR workload

usamasaqib:
### What does this PR do?

This PR adds a new workload tcp_crr. This is similar to tcp_rr but with the difference that the client initiates a new connection for every request-response.

The workload stresses the tcp connection creation code path.

### Additional Notes
PR stack
[1] https://github.com/DataDog/lading/pull/1905
[2] https://github.com/DataDog/lading/pull/1906 <-- This


lading/src/neper/flow.rs
```diff
@@ -37,28 +37,45 @@ pub(crate) struct FlowMap<S> {
...
+        }
+
+        Err(FlowMapError::NoCapacity(self.inner.capacity()))
     }
 
     /// Get a mutable reference to the flow at the given token.
     pub(crate) fn get_mut(&mut self, token: Token) -> Option<&mut Flow<S>> {
-        self.inner.get_mut(token.0).and_then(|slot| slot.as_mut())
+        let idx = token.0 % self.inner.capacity();
+        self.inner.get_mut(idx).and_then(|slot| slot.as_mut())
```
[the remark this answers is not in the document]
usamasaqib: All removes are followed up a deregister, so this is not possible. 
Since this is an invariant, adding an assert in get_mut and remove

lading/src/neper/rr.rs
```diff
@@ -284,19 +321,106 @@ fn client_thread_main(
...
+                    } else {
+                        flow.stream = new_stream;
+                        flow.state = ClientState::Connecting;
+                        flow.xfer = 0;
+                    }
+                }
+                Err(e) => {
+                    trace!("reconnect to {addr} failed: {e}");
+                    metrics.connections_failed.add(1);
+                    let _ = flows.remove(token);
```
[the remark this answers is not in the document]
usamasaqib: We partially mitigate this by increasing the local port range for all sockets created by the generator in https://github.com/DataDog/lading/pull/1906/changes/69e43c064f4de06b65aaab6d0400fa05f6823747

This is not a complete solution but addressing this edge cases add more code than is needed right now. We can evaluate later whether users actually want to create a number of flows which can cause this to even occur.

lading/src/neper/rr.rs
```diff
@@ -284,19 +425,128 @@ fn client_thread_main(
...
+    if matches!(flow.state, ClientState::Connecting) {
+        match flow.stream.take_error() {
+            Ok(None) => {
+                flow.state = ClientState::SendRequest;
+                flow.xfer = request_buf.len();
+                metrics.connections_initiated.add(1);
+                // fall through
+            }
+            Ok(Some(e)) => {
+                warn!("connect failed: {e}");
```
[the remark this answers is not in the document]
usamasaqib: We are not stress testing the receiver. Connect should not fail excessively.

lading/src/blackhole/tcp_crr.rs
```diff
@@ -0,0 +1,143 @@
...
+            data_addr: SocketAddr::new(self.config.addr, self.config.data_port),
+            control_addr: SocketAddr::new(self.config.addr, self.config.control_port),
+            threads: self.config.threads.get(),
+            flows: self.config.flows.get(),
+            request_size: self.config.request_size.get(),
+            response_size: self.config.response_size.get(),
+            no_delay: self.config.no_delay,
+            backlog: self.config.backlog,
+        };
+        rr::run_server(params, self.metric_labels, self.shutdown, "tcp_crr").await?;
```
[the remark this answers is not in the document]
preinlein: Let's double check this.

lading/src/neper/flow.rs
```diff
@@ -37,28 +37,61 @@ pub(crate) struct FlowMap<S> {
...
     /// Get a mutable reference to the flow at the given token.
+    ///
+    /// Returns `None` for an empty slot. A slot occupied by a flow with a
+    /// different token indicates a token-collision bug (tokens congruent
+    /// modulo capacity share a slot), so it is asserted against.
     pub(crate) fn get_mut(&mut self, token: Token) -> Option<&mut Flow<S>> {
-        self.inner.get_mut(token.0).and_then(|slot| slot.as_mut())
+        let idx = token.0 % self.inner.capacity();
+        let flow = self.inner.get_mut(idx).and_then(|slot| slot.as_mut())?;
+        assert_eq!(
```
preinlein, line 77: Why are we asserting in non-test code?

This is going to cause some awkward panics. Can we bubble up the error instead?

lading/src/neper/rr.rs
```diff
@@ -46,6 +46,15 @@ pub enum Error {
     Config(String),
 }
 
+/// Which neper-style protocol the client is driving.
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum Mode {
```
preinlein, line 51: Rather than set a mode and then pass that across the whole code path, can we instead define different implementations for RR vs CRR for the behaviors that differ?

We could use a trait that clearly defines where these modes differ; the methods would have different implementations and the interface on the trait would inform the next engineer what it means for a mode to be different.

I also think this partially defeats the purpose of the refactor. We extracted "common" logic out and now we're putting exceptions back in. If we end up having more "mode" specific functionality, it will be much easier to reason about it if there's a trait there already.

lading/src/neper/rr.rs
```diff
@@ -116,6 +146,7 @@ const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(5);
 ///
 /// Returns an error if configuration is invalid, the blackhole control port
 /// is never reachable, or a worker thread panics.
+#[allow(clippy::too_many_lines)]
```
preinlein, line 149: We shouldn't reach for clippy ignore comments unless we really need to. What's preventing us from breaking this function into smaller pieces?

lading/src/neper/rr.rs
```diff
@@ -231,44 +264,152 @@ pub(crate) async fn run_client(
...
+    mode: Mode,
     shutdown_flag: &AtomicBool,
     metrics: &ThreadMetrics,
 ) {
     let mut poll = Poll::new().expect("failed to create mio::Poll");
     let mut events = Events::with_capacity(num_flows as usize);
     let request_buf = vec![0u8; request_size];
     let mut response_buf = vec![0u8; response_size];
-    let mut flows: FlowMap<ClientState> = FlowMap::new();
+    let mut flows: FlowMap<ClientState> = FlowMap::new(num_flows as usize);
```
preinlein, line 385: Does it make sense to cast this to usize instead of making FlowMap only accept u16?

preinlein (commented):
With these new generators/blackholes, would it be possible to add any tests?

Could we also scan thru and make sure we do error bubbling up 🙏

lading/src/neper/flow.rs
usamasaqib: This is an invariant so this condition should never ever happen. Making it an error condition bakes it in the API as something which is possible.

This is why I am using an assert here.

lading/src/neper/rr.rs
usamasaqib: That is a good point. I can try to do that.

lading/src/neper/rr.rs
usamasaqib: I'll break it up

lading/src/neper/flow.rs
preinlein: Well, asserts are dropped in release builds so this becomes a no-op. 
We _could_ use a panic here but I do believe we're trying to drive down the number of panics to 0 for lading.

@goxberry is currently working on something to put a linting rule around `expect`: https://github.com/DataDog/lading/pull/1882 (huge stack that removes EVERY expect)

We just don't want lading to crash in unexpected ways. Having an error here bubble up doesn't change anything apart from graceful handling. You can still have the invariant be present.

If possible, it would be really neat to have some tests that exercise the invariant. This might be too difficult to do but it would be good to try.

lading/src/blackhole/tcp_crr.rs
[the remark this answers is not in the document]
usamasaqib: It is fixed here a954228de58e77ce92ed6d55dfdba6444e08f2f3

lading/src/neper/flow.rs
usamasaqib: I don't know if this standard is suitable for workloads in lading. It definitely make sense for the core, but in this case I just want a piece of software that does a very specific operation. If it cannot do it, then it should just blow up. I do not want it try to be graceful or fallback on anything.

It just seems that the effort of plumbing errors throughout the codebase is not worth it.

usamasaqib: Good to know about assert though.

preinlein: > I don't know if this standard is suitable for workloads in lading. It definitely make sense for the core, but in this case I just want a piece of software that does a very specific operation. If it cannot do it, then it should just blow up. I do not want it try to be graceful or fallback on anything.
> 
> It just seems that the effort of plumbing errors throughout the codebase is not worth it.

What makes it unsuitable for these operations/generators/blackholes? 

From what I've seen, you have a very clear entry point to the generator/blackhole execution that already often already follows the `Result<(), Error>` interface. What makes calling the logic inside of these not suitable for bubbling errors up?

usamasaqib: Suitable may not be the right word here I guess. What I was trying to point out was that as per my understanding this is just a stylistic concern? The panics are irrecoverable conditions as far as I am concerned so they should never happen. There will not be any code in lading for handling these errors either.

But if it is going to be a project wide lint, then I can remove these cases.

preinlein: It's a bit stylistic but it's also that in general, chaotically exiting the program is not desirable. Even if "it should never happen", doesn't mean it will unless we guarantee the invariants. Any future code change, could expose this code path as problematic without us realizing. 

And then if we contrast graceful handling vs a panic.
- If the panic is in the wrong place, Tokio can just swallow it. The program will continue to execute but in a weird state.
- If we panic, it means normal exit code (any kind of cleanup/post execution code) won't run. That could lead the system to being in an expected state and it could make debugging the problem much harder.

Vs gracefully handling that error and exiting lading, we'd see `lading` exit with a failure/error code but at least we have guarantees that it does exit and it does cleanup after itself.

---

And then personally, I think that we need to be cognizant of the unknown unknowns - we don't know what we don't know. There's a possibility where we realize there are conditions where this invariant does not hold and/or there _could_ be ways to recover. By wiring up error bubbling, it makes it so that the structure is in place already. 

I would much prefer to realize "oh cool, there's something I need to do here and there's already error handling in place and I can add the logic there" vs "oh shoot, error bubbling isn't done, I need to add that first and add the error handling at the right place to fix the issue".

--- 
Having said all that, I'm sorry I dinged you on this one. We have traditionally "allowed" panics in `lading` code and it's just recently that we've determined that this is undesirable and we shouldn't be allowing that.

lading/src/generator/tcp_crr.rs
```diff
@@ -0,0 +1,131 @@
+//! TCP connect/request/response (`tcp_crr`) generator — the client side.
```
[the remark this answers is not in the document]
preinlein: Let's double-check this and if so, let's address it.

lading/src/generator/tcp_crr.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Run the generator to completion or until a shutdown signal is received.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if a worker thread panics or configuration is invalid.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `addr` cannot be parsed as an IP address.
```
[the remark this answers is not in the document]
preinlein: @usamasaqib 

I'm curious, why did we downvote this? We don't currently have any validation in place for `addr` and it can very easily be malformed. Why don't we return an error if we can't parse it?

lading/src/neper/flow.rs
preinlein: For now, let's drop the asserts and use an explicit expert/fatal.

preinlein (commented):
I started reviewing this and realized you might not want a review on this yet. I'll leave my current comments and stop the review here. Let me know if & when you want me to continue the review.

PR #1908, opened 2026-07-07
feat(otlp): extend metric testing support  

lucastemb:
### What does this PR do?

Adds support for the remaining OTLP metrics (Exponential Histogram, Histogram, and Summary) in Agent Data Plane correctness tests. 

### Motivation

Will be used to test whether these metric types are being properly handled by ADP in the Saluki repo when compared to the Agent. 

**Supports:**
- #https://github.com/DataDog/saluki/pull/1319
### Related issues

for this Pull Request.

### Additional Notes

Listed below is some additional information for the metrics added (how they are constructed and how they are expected to behave): 

- [Histogram](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram)
- [Exponential Histogram](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#exponentialhistogram)
- [Summary](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#summary-legacy)

### DIFF
--- a/lading_payload/src/opentelemetry/metric.rs
+++ b/lading_payload/src/opentelemetry/metric.rs
@@ -133,6 +133,12 @@ where
     point.max = Some(max);
 }
 
+fn set_histogram_sum_at_least_max(point: &mut HistogramDataPoint) {
+    if let (Some(sum), Some(max)) = (&mut point.sum, point.max) {
+        *sum = sum.max(max);
+    }
+}
+
 fn exp_histogram_bucket_bounds(scale: i32, index: i32) -> (f64, f64) {
     let base = 2.0_f64.powf(2.0_f64.powi(-scale));
     (base.powi(index), base.powi(index + 1))
@@ -178,6 +184,12 @@ where
 
     point.min = Some(min.min(max));
     point.max = Some(max.max(min));
+}
+
+fn set_exp_histogram_sum_at_least_max(point: &mut ExponentialHistogramDataPoint) {
+    if let (Some(sum), Some(max)) = (&mut point.sum, point.max) {
+        *sum = sum.max(max);
+    }
 }
 
 /// Configure the OpenTelemetry metric payload.
@@ -595,6 +607,12 @@ impl<'a> SizedGenerator<'a> for OpentelemetryMetrics {
                                     {
                                         *current = current.min(previous);
                                     }
+                                    if let Some((_, last)) =
+                                        populated_bucket_range(&point.bucket_counts)
+                                    {
+                                        let (_, upper) = histogram_bucket_bounds(point, last);
+                                        point.max = Some(upper + self.incr_f);
+                                    }
                                     if let (Some(previous), Some(current)) =
                                         (previous_max, point.max.as_mut())
                                     {
@@ -603,6 +621,7 @@ impl<'a> SizedGenerator<'a> for OpentelemetryMetrics {
                                     if let Some(sum) = &mut point.sum {
                                         *sum += self.incr_f;
                                     }
+                                    set_histogram_sum_at_least_max(point);
                                 } else {
                                     // Delta: fresh observations each window.
                                     // Use 1..=10 (not 0..=10) so count stays
@@ -615,6 +634,7 @@ impl<'a> SizedGenerator<'a> for OpentelemetryMetrics {
                                     if let Some(sum) = &mut point.sum {
                                         *sum = rng.random_range(1.0_f64..=1000.0);
                                     }
+                                    set_histogram_sum_at_least_max(point);
                                 }
                             }
                         }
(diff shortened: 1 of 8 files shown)


webern (commented):
.

lading_payload/src/opentelemetry/metric.rs
```diff
@@ -387,45 +536,164 @@ impl<'a> SizedGenerator<'a> for OpentelemetryMetrics {
...
+                                    // Delta: fresh observations each window.
+                                    // Use 1..=10 (not 0..=10) so count stays
+                                    // non-zero and the encoded size is stable.
+                                    for bc in &mut point.bucket_counts {
+                                        *bc = rng.random_range(1_u64..=10);
+                                    }
+                                    point.count = point.bucket_counts.iter().sum();
+                                    set_histogram_min_max(point, rng);
+                                    if let Some(sum) = &mut point.sum {
+                                        *sum = rng.random_range(1.0_f64..=1000.0);
```
webern, line 635: `P0?`

Seems legit. Sounds like it's saying the clamping behavior could lead to an inconsistency between the bucket sum and the bucket values.

> Could we derive this `sum` from the generated bucket distribution, or otherwise constrain it to be consistent with the buckets? The explicit bounds are positive and every delta bucket is populated, so `max` can be well above `1000` while `sum` is capped at `1000`, producing a histogram with `sum < max`. The cumulative path has the same issue: every bucket count grows by `incr_i`, but `sum` only grows by one `incr_f` per tick.

lading_payload/src/opentelemetry/metric.rs
```diff
@@ -1033,6 +1320,196 @@ mod test {
...
+                    attributes_per_scope: ConfRange::Constant(0),
+                    metrics_per_scope: ConfRange::Constant(8),
+                    attributes_per_metric: ConfRange::Constant(0),
+                },
+                metric_weights,
+            };
+            let mut budget = 1_000_000;
+            let mut rng = SmallRng::seed_from_u64(42);
+            let mut otel_metrics = OpentelemetryMetrics::new(config, budget, &mut rng)?;
+            let resource_metric = otel_metrics.generate(&mut rng, &mut budget)?;
```
webern, line 1512: It noted that you only check one generate call, and that adding a second generate call would further assert the invariants necessary for cumulative metrics.

**AI Finding:**

Could we extend this test to exercise cumulative Histograms and Exponential Histograms across at least two calls to `generate()`? The current test calls `generate()` only once, so it verifies that one snapshot is well-formed but cannot catch a regression that resets cumulative state or updates `min`/`max` in the wrong direction.

With a cumulative-only configuration, could we retain the same generator and compare the corresponding points from successive calls, roughly:

```rust
let first = generator.generate(&mut rng, &mut budget)?;
let second = generator.generate(&mut rng, &mut budget)?;

assert!(second.count >= first.count);
assert!(second.sum >= first.sum);
assert!(second.min <= first.min);
assert!(second.max >= first.max);
```

The test should also verify that each Histogram bucket count is non-decreasing, and that the zero, positive, and negative bucket counts are non-decreasing for Exponential Histograms. (The exact comparisons may need to account for these fields being optional.) This would catch cumulative metrics accidentally behaving like delta metrics while each individual snapshot still appears valid.

lading_payload/src/opentelemetry/metric/templates.rs
```diff
@@ -563,7 +789,13 @@ mod test {
                             _ => panic!("invalid aggregation temporality"),
                         }
                     }
-                    _ => panic!("invalid metric data"),
+                    Data::Histogram(_) => assert!(config.metric_weights.histogram_delta >= 1
```
webern, line 793: This is just a nit/optional: I was looking for testing gaps and found this:

> Could we test the exact temporality mapping for the new types here? For example, a histogram-delta-only configuration should always produce `aggregation_temporality == 1`, while a histogram-cumulative-only configuration should produce `2` (and likewise for exponential histograms). These values select the mutation branch in `metric.rs`, but the current assertions only verify that some weight is nonzero.

lading_payload/src/opentelemetry/metric.rs
```diff
@@ -1033,6 +1320,196 @@ mod test {
...
+
+            for scope_metric in &resource_metric.scope_metrics {
+                for metric in &scope_metric.metrics {
+                    match &metric.data {
+                        Some(Data::Histogram(histogram)) => {
+                            for point in &histogram.data_points {
+                                assert!(point.min.is_some());
+                                assert!(point.max.is_some());
+                                assert!(point.min <= point.max);
```
webern, line 1522: Seems like it caught an inconsistency in the assertions with a simple fix:

> Could we also assert `point.count == point.bucket_counts.iter().sum()` for explicit histograms? The equivalent invariant is checked for exponential histograms below, but not for this branch, even though `generate()` recomputes `count` from the bucket counts.

webern (approved):
Seems like nits except maybe for `P0?`.

PR #1911, opened 2026-07-15
feat(lading): add record policy to allowlist recorded series in datadog blackhole

armleth:
Adds an optional `record` policy to the `datadog` blackhole config, controlling which received series are recorded as lading capture metrics:

- `all` (default): record every series with every tag. Unchanged behaviour.
- `disabled`: record no series; payloads are still decoded and bytes still counted.
- `series_to_keep: [...]`: record only series whose metric name is listed (with all their tags); drop the rest.

```yaml
blackhole:
  - datadog:
      v2:
        binding_addr: "0.0.0.0:9091"
      record:
        series_to_keep: [my.metric.one, my.metric.two]
```

An allowlist by series name keeps the recorded outputs explicit in the lading config and avoids accidentally recording (and forwarding to org2) generator/fuzz-produced series.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 - Datadog blackhole now accepts a `record` policy (`all` / `disabled` /
   `tags_to_drop: [...]`) controlling which received series are recorded as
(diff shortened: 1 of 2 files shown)


GeorgeHahn (approved):
This looks great, thanks for the contribution! ~~The cargo-deny failure is unrelated; I will put a PR up to fix it shortly.~~ The cargo-deny failure was fixed in a PR that just merged and should go away if you rebase.

PR #1926, opened 2026-08-06
Namespace datadog blackhole target metrics & default off

GeorgeHahn:
Two breaking changes:

- Apply a prefix to target metrics received by the datadog blackhole for consistency with other target metrics collectors.
- Disable metrics recording by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+
+## Changed
+- **Breaking Change**: Datadog blackhole now records received series under the
+  `target/` prefix, matching the prometheus and expvar target metrics
+  collectors. The `record` policy still matches on the unprefixed series name.
+- **Breaking Change**: Datadog blackhole does not record any target metrics by
+  default.
+
+## Added
 - Datadog blackhole now accepts a `record` policy (`all` / `disabled` /
   `series_to_keep: [...]`) controlling which received series are recorded as
   capture metrics.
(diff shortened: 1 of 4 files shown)


lading/src/blackhole/datadog.rs
```diff
@@ -49,6 +49,9 @@ use tracing::{debug, error, info, trace, warn};
 use super::General;
 use crate::proto::datadog::intake::metrics::MetricPayload;
 
+/// Namespace for incoming metrics; matches other target-originated metrics.
+const TARGET_PREFIX: &str = "target/";
```
preinlein, line 53: Does this imply that we'll have a metric rename happening? 

My concern is with the variety of dashboards/tools/scripts that already exist that use these.

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+
+## Changed
+- **Breaking Change**: Datadog blackhole now records received series under the
```
preinlein, line 10: Ok, that answers my previous question. 

Going to approve but we should let the team/users know when we do the release.

preinlein (approved):
There's no **clean** way to double-emit and do a "migration" with our users. We'll have to let this be a breaking change, make comms, and let our on-call know.

lading/src/blackhole/datadog.rs
GeorgeHahn: Yes for experiments that use the datadog blackhole. Those experiments will also need adjustment to explicitly opt into target metric collection.

CHANGELOG.md
GeorgeHahn: Absolutely yeah. I'll land the experiment changes along with teams' update PRs and keep the breaking notes in the public announcement.

PR #1929, opened 2026-09-11
feat(payload): v1.0 (idx/ETP) trace agent payload - hardening, wire-format validation, and fingerprint corpus

ajgajg1134:
## Summary

Add trace-agent payload generation for `/v1.0/traces` (idx/ETP), with service-graph workloads, error injection, span links, and events. `chunks_per_payload` is an exact batch size: a budget too small for the complete batch rejects it without dropping or adding chunks. Cache construction fails with an actionable error after 1024 consecutive rejected blocks.

Preserve configured error attributes, validate complete tracer captures and generated wire output, and separate configuration, encoding, and tests for review.

## Notes for reviewers

Start with `lading_payload/src/trace_agent/v1/README.md` for batching and workload constraints. `config.rs` contains validation, `encoder.rs` contains the wire schema, `v1.rs` contains generation, and `test.rs`/`golden.rs` contain the invariants and fixture checks.

The wire format is idx/ETP MessagePack for `/v1.0/traces`: integer-keyed payload/chunk/span maps, a payload-scoped streaming string table (empty string at index zero), and flat key/type/value attribute triples. Reference implementations are the [Agent idx encoder/decoder](https://github.com/DataDog/datadog-agent/tree/6e10d6cf0e1ef764bfe4390a68c55cee415d264c/pkg/proto/pbgo/trace/idx), [Saluki decoder](https://github.com/DataDog/saluki/tree/0af0bdff879c6809985b4e5ddddff31667632c61/lib/saluki-components/src/decoders/datadog), and [dd-trace-go encoder](https://github.com/DataDog/dd-trace-go/blob/bf08f5f9aafed5797e3666fbdd2325caa57e3704/ddtrace/tracer/payload_v1.go).

The golden request bodies came from `apm-v1-trace-smoke` using dd-trace-go v2.11.0-dev.1. Its discovery shim advertised v1.0 and native span events and forwarded requests to a recording sink. The fixtures cover nested services, errors, links, and scalar event attributes.

The test-only decoder checks schema-level meaning. Golden round trips include every chunk of both captures, including errors and cross-chunk string references. Generated-payload tests verify exact batch sizes, span counts, errors, links/events, and determinism. Equivalent encodings can differ in attribute order, integer widths, and therefore streaming string indices, so raw byte equality is not a replacement for structural comparison. Golden tests are not a substitute for an ingestion check against the real agent.

The service graph is independent of the OTLP generator: explicit names/resources/types, first-service roots, static scalar attributes, and no dictionary pools, per-call attributes, `max_repeat`, or explicit root selection. Port topology and rates semantically; do not assume matching attribute cardinality. Shared traversal/configuration can be considered separately from this encoder.

The v1 fingerprint intentionally changes with exact batching: single-chunk bodies repeat metadata and restart the string table, and geometric sizing probes no longer consume RNG draws. Corpus entropy changes from 7.0414 to 6.6439 bits/byte; the other 12 fingerprint baselines are unchanged.

## Validation

- [x] `ci/fmt`
- [x] `ci/custom_lints` (pinned ast-grep 0.39.5)
- [x] `cargo clippy -p lading-payload --all-targets --all-features`
- [x] `cargo nextest run --workspace --exclude sheepdog` (508 passed)
- [x] `ci/fingerprint` (13 passed; v1 fingerprint also verified reproducible)
- [x] New example passes `lading config-check`
- [x] Impossible 100-chunk / 1 KiB batch returns an actionable error through the rebuilt payload tool
- [ ] Full `ci/validate`: shellcheck and formatting pass, but all-feature compilation is blocked locally by missing macOS FUSE. Workspace tests above ran without optional FUSE.


lading_payload/src/trace_agent/v1/encoder.rs
```diff
@@ -0,0 +1,413 @@
+//! Streaming string-table `MessagePack` encoder for v1.0 payloads.
+use std::io::Write;
+
+use rmp::encode::ValueWriteError;
+use rustc_hash::FxHashMap;
+
+use super::{AttributeValue, Span, SpanEvent, SpanLink, TraceChunk, TracerPayload};
+use crate::Error;
```
blt: General project preference, give encoder.rs its own Error, then use thiserror `#[from]` to roll that up into a catch-all error. That way you can have more precise control over encoder errors -- and avoid stringy errors -- and not have churn somewhere else in the project if you're not handling specifically `Error::Encoder(_)`.

lading_payload/src/trace_agent/v1/encoder.rs
```diff
@@ -0,0 +1,413 @@
...
+fn write_map_len<W: Write>(w: &mut W, len: usize) -> Result<(), Error> {
+    rmp::encode::write_map_len(w, u32::try_from(len).map_err(cast_error)?)
+        .map(|_| ())
+        .map_err(map_write_error)
+}
+
+fn write_array_len<W: Write>(w: &mut W, len: usize) -> Result<(), Error> {
+    rmp::encode::write_array_len(w, u32::try_from(len).map_err(cast_error)?)
+        .map(|_| ())
+        .map_err(map_write_error)
```
blt: Just picking one of these `map_err` instances out of the group, but consistent with my comment earlier I think you could collapse this neatly to a ? and avoid some of the repetition here.

lading_payload/src/trace_agent/v1/encoder.rs
```diff
@@ -0,0 +1,413 @@
...
+    /// Encodes the payload into its `v1.0` `MessagePack` representation.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying writer fails. Writing into an in-memory buffer cannot
+    /// fail, so callers that do so can treat this as infallible.
+    pub(super) fn encode(&self) -> Result<Vec<u8>, Error> {
+        let mut buf = Vec::new();
+        let mut strings = StringTable::new();
+        write_tracer_payload(&mut buf, &mut strings, self)?;
```
blt: I think this code would be more clear if you, instead of having bare C-style functions that have to pass their state around, had TracerPayload have private methods that do essentially what you have now, just collapsed in under the struct.

lading_payload/src/trace_agent/v1/golden.rs
```diff
@@ -0,0 +1,607 @@
+//! Golden tests for the v1.0 wire format, validated against a real tracer payload.
```
blt, line 1: Oh, novel for the project. I like it.

lading_payload/src/trace_agent/v1/golden.rs
```diff
@@ -0,0 +1,607 @@
...
+
+    fn double(&mut self) -> Result<f64, String> {
+        rmp::decode::read_f64(&mut self.data).map_err(read_error)
+    }
+
+    fn inline_str(&mut self) -> Result<String, String> {
+        let len = rmp::decode::read_str_len(&mut self.data).map_err(read_error)?;
+        let bytes = self.take(len as usize)?;
+        String::from_utf8(bytes.to_vec()).map_err(|e| format!("invalid UTF-8: {e}"))
+    }
```
blt: Ah I was concerned when I saw read_error. Instead of the error type here being a String prefer a derivation from `thiserror::Error`.

lading_payload/src/trace_agent/v1/README.md
```diff
@@ -0,0 +1,83 @@
+# v1.0 trace payloads
```
blt, line 1: Nice, thank you for dropping a README here.

lading_payload/src/trace_agent/v1.rs
```diff
@@ -0,0 +1,553 @@
...
+/// A single span.
+///
+/// Strings and attributes are reference-counted handles into the resolved operation graph, so a
+/// span is assembled by cloning cheap handles rather than re-allocating its fields.
+#[allow(clippy::struct_field_names)]
+#[derive(Clone, Debug, Default, PartialEq)]
+struct Span {
+    service: Arc<str>,
+    name: Arc<str>,
+    resource: Arc<str>,
```
blt: Hmm could these be Rc and etc below? It's possible the ownership does need the cross-task Arc but just asking to reconfirm.

lading_payload/Cargo.toml
```diff
@@ -20,6 +20,7 @@ opentelemetry-proto = { workspace = true }
 prost = { workspace = true }
 rand = { workspace = true, features = ["std", "std_rng"] }
 rustc-hash = { workspace = true }
+rmp = { version = "0.8", default-features = false }
```
blt: Project has a preference for all dependencies -- also for base64 below -- to be in root Cargo.toml, if you would.

Cargo.toml
```diff
@@ -14,6 +14,7 @@ members = [
 [workspace.dependencies]
 anyhow = "1.0"
 arc-swap = { version = "1.8" }
+base64 = "0.22"
```
blt, line 17: My bad, I really should write a lint for this. Would you keep this to the form `{ version = "0.22", default-features = false }` and then flag on whatever features you need at the point of use?

blt (approved):

lading_payload/src/block.rs
```diff
@@ -566,6 +569,35 @@ impl Cache {
...
+    max_block_size: u32,
+) -> Result<Block, SpinError>
+where
+    S: crate::Serialize,
+    R: Rng + ?Sized,
+{
+    match construct_block(rng, serializer, max_block_size) {
+        Ok(block) => Ok(block),
+        Err(SpinError::EmptyBlock) => Err(SpinError::InvalidConfig(format!(
+            "No payload fit after 1024 consecutive attempts, including a direct attempt at maximum_block_size={max_block_size} bytes; increase maximum_block_size or reduce the payload size (for v1.0, chunks_per_payload or the service graph)."
```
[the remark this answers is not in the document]
ajgajg1134: While extremely unlikely, this is easy to fix so I've added a small retry loop here to make this basically impossible.

